PR #9, opened 2021-06-02
Update Discord Version to 0.0.15

javaarchive:
Discord sometimes can't auto update on linux due to having to change files in directories outside the home directory which it can't because of permissions. Recently version 0.0.15 was released.


wizard-28 (approved):
Okay merging it

PR #17, opened 2021-06-06
Dunst exist in universe so it might be installed

kaistian:


kaistian:
I guess it's failing because of this bug https://github.com/pacstall/pacstall/pull/53

wizard-28 (approved):
Okay, but @pacstall/dunst needs to review it

Surendrajat (approved):

PR #14, opened 2021-06-04
Update Soundux to 0.2.7

D3SOX:

PR #19, opened 2021-06-06
Soundux: Remove obsolete dependency `pulseaudio-utils`

D3SOX:
We no longer require this dependency and it can safely be removed (forgot to do this in the 0.2.7 update PR, sorry)


packages/soundux/soundux.pacscript
```diff
@@ -2,7 +2,7 @@ name="soundux"
 version="0.2.7"
 url="https://github.com/Soundux/Soundux/releases/download/0.2.7/soundux-0.2.7.tar.gz"
 build_depends="build-essential cmake libx11-dev libxi-dev libwebkit2gtk-4.0-dev libappindicator3-dev libssl-dev libpulse-dev"
```
wizard-28, line 4:
```suggestion
build_depends="cmake libx11-dev libxi-dev libwebkit2gtk-4.0-dev libappindicator3-dev libssl-dev libpulse-dev"
```
`build-essential` is obsolete too, as it is a pacstall dependency, no need to add it separately

Citation
https://github.com/pacstall/pacstall/blob/35b3f057d4104fa6ded75e8faecc923e1ee8477d/install.sh#L91

wizard-28 (changes requested):

packages/soundux/soundux.pacscript
D3SOX: Thanks!

wizard-28 (commented):
Are you sure that this works without a `prepare()` function?

I tried removing the `prepare()` function from another pacscript, and I got an error.

Citation (Pacstall currently executes `prepare` regardless if it exists or not.)
https://github.com/pacstall/pacstall/blob/35b3f057d4104fa6ded75e8faecc923e1ee8477d/misc/scripts/install-local.sh#L198

D3SOX:
> Are you sure that this works without a `prepare()` function?
> 
> I tried removing the `prepare()` function from another pacscript, and I got an error.
> 
> Error I got
> 
> ```
> /usr/share/pacstall/scripts/install-local.sh: line 199: prepare: command not found
> ```
> 
> Citation (Pacstall currently executes `prepare` regardless if it exists or not.)
> https://github.com/pacstall/pacstall/blob/35b3f057d4104fa6ded75e8faecc923e1ee8477d/misc/scripts/install-local.sh#L198

Hm I tested it a VM and the current package installed successfully. I removed the `prepare` in #14 because I thought it's unnecessary

wizard-28:
> > Are you sure that this works without a `prepare()` function?
> > I tried removing the `prepare()` function from another pacscript, and I got an error.
> > Error I got
> > ```
> > /usr/share/pacstall/scripts/install-local.sh: line 199: prepare: command not found
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Citation (Pacstall currently executes `prepare` regardless if it exists or not.)
> > https://github.com/pacstall/pacstall/blob/35b3f057d4104fa6ded75e8faecc923e1ee8477d/misc/scripts/install-local.sh#L198
> 
> Hm I tested it a VM and the current package installed successfully. I removed the `prepare` in #14 because I thought it's unnecessary. Which version of pacstall are you using?

Latest develop branch. See the citation

wizard-28:
> > Are you sure that this works without a `prepare()` function?
> > I tried removing the `prepare()` function from another pacscript, and I got an error.
> > Error I got
> > ```
> > /usr/share/pacstall/scripts/install-local.sh: line 199: prepare: command not found
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Citation (Pacstall currently executes `prepare` regardless if it exists or not.)
> > https://github.com/pacstall/pacstall/blob/35b3f057d4104fa6ded75e8faecc923e1ee8477d/misc/scripts/install-local.sh#L198
> 
> Hm I tested it a VM and the current package installed successfully. I removed the `prepare` in #14 because I thought it's unnecessary. Which version of pacstall are you using?

The package installed successfully but it also prints that error... Which looks bad

packages/soundux/soundux.pacscript
```diff
@@ -1,8 +1,8 @@
 name="soundux"
 version="0.2.7"
 url="https://github.com/Soundux/Soundux/releases/download/0.2.7/soundux-0.2.7.tar.gz"
```
wizard-28, line 3:
```suggestion
url="https://github.com/Soundux/Soundux/releases/download/"$version"/soundux-"$version".tar.gz"
```
Reuse the `version` variable

wizard-28 (changes requested):

D3SOX:
> Latest develop branch

Yeah I've used the latest release. I didn't know this is needed.

D3SOX:
> but it also prints that error... Which looks bad

I think the prepare function should be optional. I added it for now but this should be resolved in pacstall.

packages/soundux/soundux.pacscript
```diff
@@ -1,13 +1,17 @@
 name="soundux"
 version="0.2.7"
-url="https://github.com/Soundux/Soundux/releases/download/"$version"/soundux-"$version".tar.gz"
```
wizard-28: Why remove the `"`? It prevents word globbing and other weird things..

D3SOX: It's already in quotes

wizard-28: Hmm... ok then.

wizard-28 (approved):
Good to go!

PR #22, opened 2021-06-07
[test] Add shebang to the pacscripts I maintain

wizard-28:
This adds shebangs to my pacscripts


Henryws:
Why?

wizard-28:
CodeFactor can't check for errors if the file has no `.sh` extension

So I saw this issue, and it said that they have added checks for shebang, so I am testing my scripts with shebang to see whether code factor picks it up

https://github.com/codefactor-io/codefactor/issues/2#issuecomment-623287639

Henryws:
I think that the ci we use is fine to check the script and pacscripts have a standardized format that's so simple, there's no way someone's could mess it up

wizard-28:
The current CI is broken, it doesn't even execute properly for most of the pacscripts xD

Henryws:
It does, but only on one's where people submit a single package per pr

wizard-28:
Umm, looks like CF isn't detecting the file based on the shebang, well closing this PR

wizard-28:
> It does, but only on one's where people submit a single package per pr

It doesn't when the pacscript uses `replace`

PR #39, opened 2021-06-12
Update Soundux pacscript

D3SOX:
This now installs the doc, license and a binary symlink that it can be started via `soundux` in the terminal

### DIFF
--- a/packages/soundux/soundux.pacscript
+++ b/packages/soundux/soundux.pacscript
@@ -23,11 +23,11 @@ build() {
 install() {
   sudo make install DESTDIR="$pkgdir"
   # install binary symlink
-  mkdir -p "$pkgdir/usr/bin/"
-  ln -sf /opt/soundux/soundux "$pkgdir/usr/bin/soundux"
+  sudo mkdir -p "$pkgdir/usr/bin/"
+  sudo ln -sf /opt/soundux/soundux "$pkgdir/usr/bin/soundux"
   
   # install doc
-  install -Dm 644 -t "$pkgdir/usr/share/doc/$name" "../README.md"
+  sudo install -Dm 644 -t "$pkgdir/usr/share/doc/$name" "../README.md"
   # install license
-  install -Dm 644 -t "$pkgdir/usr/share/licenses/$name" "../LICENSE"
+  sudo install -Dm 644 -t "$pkgdir/usr/share/licenses/$name" "../LICENSE"
 }
--- a/packages/soundux/soundux.pacscript
+++ b/packages/soundux/soundux.pacscript
@@ -1,4 +1,5 @@
 name="soundux"
+pkgdir="/usr/src/pacstall/$name"
 version="0.2.7"
 url="https://github.com/Soundux/Soundux/releases/download/$version/soundux-$version.tar.gz"
 build_depends="cmake libx11-dev libxi-dev libwebkit2gtk-4.0-dev libappindicator3-dev libssl-dev libpulse-dev"
@@ -20,5 +21,13 @@ build() {
 }
 
 install() {
-  sudo make install DESTDIR=/usr/src/pacstall/soundux
+  sudo make install DESTDIR="$pkgdir"
+  # install binary symlink
+  mkdir -p "$pkgdir/usr/bin/"
+  ln -sf /opt/soundux/soundux "$pkgdir/usr/bin/soundux"
+  
+  # install doc
+  install -Dm 644 -t "$pkgdir/usr/share/doc/$name" "../README.md"
+  # install license
+  install -Dm 644 -t "$pkgdir/usr/share/licenses/$name" "../LICENSE"
 }
--- a/packages/soundux/soundux.pacscript
+++ b/packages/soundux/soundux.pacscript
@@ -9,7 +9,7 @@ maintainer="D3SOX <d3sox@protonmail.com>"
 removescript="yes"
 
 _ubuntuver=$(grep "VERSION_ID=" /etc/os-release | grep -o "[0-9]*" | tr -d '[:space:]')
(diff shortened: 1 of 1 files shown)


D3SOX:
Seems like it doesn't like absolute symlinks. Is there another to create that symlink or should I remove it?

packages/soundux/soundux.pacscript
```diff
@@ -20,5 +21,13 @@ build() {
 }
 
 install() {
-  sudo make install DESTDIR=/usr/src/pacstall/soundux
+  sudo make install DESTDIR="$pkgdir"
+  # install binary symlink
+  sudo mkdir -p "$pkgdir/usr/bin/"
+  sudo ln -sf /opt/soundux/soundux "$pkgdir/usr/bin/soundux"
```
wizard-28, line 27: >  Seems like it doesn't like absolute symlinks. Is there another to create that symlink or should I remove it?

Why are you symlinking this? What is the reason?

D3SOX: I explained this in the description. That it is in the PATH.

wizard-28 (approved):
Approved!

D3SOX:
Why did you approve it? It seems like it doesn't install right now

packages/soundux/soundux.pacscript
wizard-28, line 27: Sorry for that weird approval, I did it by mistake.

So you are symlinking the binary from `/opt/` to the `pkgdir`?

Shouldn't this be the other way round? I mean the `/opt/soundux/soundux` is a non existent file

wizard-28 (changes requested):

D3SOX:
Also one other problematic thing: We don't need the pacstall dependency on Ubuntu 21.04. There is currently no option either to install a package as an apt dependency and when it fails the pacstall one.
I could remove the pacdep for the time being and do something like `apt install -y libpipewire-0.3-dev | sudo pacstall -I -y pipewire` in prepare

packages/soundux/soundux.pacscript
D3SOX: With ln the first parameter is the location where it should point and the second one where the symlink should live

wizard-28: Move the symlink to `postinst()`, then `stow` won't interfere with anything

wizard-28:
> Also one other problematic thing: We don't need the pacstall dependency on Ubuntu 21.04. There is currently no option either to install a package as an apt dependency and when it fails the pacstall one.
> I could remove the pacdep for the time being and do something like `apt install -y libpipewire-0.3-dev | sudo pacstall -I -y pipewire` in prepare

I think you mean *pipewire* dependency.

I think you can grep `/etc/os-release`'s `VERSION`, then write code that either executes `pacdeps` or `apt`


Something like this
```shell
if ! grep "21.04" /etc/os-release; then
   pacdeps=("pipewire")
fi
```

This will execute `pacdeps` only if `21.04` is not found in `/etc/os-release`

D3SOX:
Thank you that's a good idea but I think I need to tinker a bit with this because 20.10 also needs the pacdep

D3SOX:
I decided to remove that symlink for now. This is now ready for review.

packages/soundux/soundux.pacscript
```diff
@@ -20,5 +27,10 @@ build() {
...
 
 install() {
-  sudo make install DESTDIR=/usr/src/pacstall/soundux
+  sudo make install DESTDIR="$pkgdir"
+  
+  # install doc
+  sudo install -Dm 644 -t "$pkgdir/usr/share/doc/$name" "../README.md"
+  # install license
+  sudo install -Dm 644 -t "$pkgdir/usr/share/licenses/$name" "../LICENSE"
 }
```
wizard-28, line 36:
```suggestion
}

removescript() {
    sudo rm -rf "/usr/share/doc/$name"
    sudo rm -rf "/usr/share/licenses/$name"
}
```

You have to remove the empty directories after uninstallation

packages/soundux/soundux.pacscript
```diff
@@ -1,12 +1,19 @@
 name="soundux"
+pkgdir="/usr/src/pacstall/$name"
 version="0.2.7"
 url="https://github.com/Soundux/Soundux/releases/download/$version/soundux-$version.tar.gz"
 build_depends="cmake libx11-dev libxi-dev libwebkit2gtk-4.0-dev libappindicator3-dev libssl-dev libpulse-dev"
-depends="libgtk-3-0 libwnck-3-0 libappindicator3-1 libwebkit2gtk-4.0-37 libx11-6 pulseaudio"
 description="A cross-platform soundboard"
 hash="017003fc96f49df30575975f3904c0d8a500e325a9d2bca8c3dc69fed0cab0a7"
 maintainer="D3SOX <d3sox@protonmail.com>"
```
wizard-28, line 8:
```suggestion
maintainer="D3SOX <d3sox@protonmail.com>"
removescript="yes"
```

wizard-28 (changes requested):

packages/soundux/soundux.pacscript
```diff
@@ -1,12 +1,20 @@
...
-pacdeps=("pipewire")
+removescript="yes"
+
+_ubuntuver=$(grep "VERSION_ID=" /etc/os-release | grep -o "[0-9]*" | tr -d '[:space:]')
+if [ $_ubuntuver -ge 2010 ]; then
```
wizard-28, line 12:
```suggestion
if [ "$_ubuntuver" -ge 2010 ]; then
```

wizard-28 (changes requested):

wizard-28 (approved):
Approved

PR #41, opened 2021-06-14
[Update Package] Updated mutt-wizard to install license

JonBoyleCoding:


JonBoyleCoding:
I'm not sure if there is some form of release version like in AUR as there hasn't been an update in actual program version, but a change in what gets installed.

wizard-28 (approved):
Approved, so should I merge?

wizard-28:
> I'm not sure if there is some form of release version like in AUR as there hasn't been an update in actual program version, but a change in what gets installed.

You mean `pkgrel`? No it doesn't exist in pacstall

JonBoyleCoding:
> > I'm not sure if there is some form of release version like in AUR as there hasn't been an update in actual program version, but a change in what gets installed.
> 
> You mean `pkgrel`? No it doesn't exist in pacstall

Indeed, that is what I meant.

JonBoyleCoding:
> Approved, so should I merge?

If you're happy with it, I am.

wizard-28 (changes requested):
Never mind! The makefile already does this

wizard-28 (approved):
Merging

PR #50, opened 2021-06-20
updated pacget's pacscript

erikLundstedt:
it now gets the version from git for better updates
i also moved everything to pacget-git despite that being the only version
and i updated the packagelist to reflect the changes


wizard-28 (changes requested):
I don't thing you should remove the `-git` suffix, as people who download the `-git` version actually understand that they are installing the `-git` version (and the pros/cons with it, like more unstable/frequent updates). So I think removing the suffix is kind of hiding that fact.

---

> i also moved everything to pacget-git as despite that being the only version

There is a fetching software called [fastfetch](https://github.com/LinusDierheimer/fastfetch), which has it's AUR PKGBUILD script with the suffix `-git` even though it's their only way of distributing it to arch users.
*PS: I maintain their [pacscript version](https://github.com/pacstall/pacstall-programs/blob/master/packages/fastfetch-git/fastfetch-git.pacscript), and I also named it with the `-git` suffix*

erikLundstedt:
> I don't thing you should remove the `-git` suffix, as people who download the `-git` version actually understand that they are installing the `-git` version (and the pros/cons with it, like more unstable/frequent updates). So I think removing the suffix is kind of hiding that fact.
> 
> > i also moved everything to pacget-git as despite that being the only version
> 
> There is a fetching software called [fastfetch](https://github.com/LinusDierheimer/fastfetch), which has it's AUR PKGBUILD script with the suffix `-git` even though it's their only way of distributing it to arch users.
> _PS: I maintain their [pacscript version](https://github.com/pacstall/pacstall-programs/blob/master/packages/fastfetch-git/fastfetch-git.pacscript), and I also named it with the `-git` suffix_

I added the `-git` suffix I didn't remove it

packages/pacget-git/pacget-git.pacscript
```diff
@@ -0,0 +1,44 @@
...
+name="pacget-git"
+pkgname="pacget"
+pkgdir="/usr/src/pacstall/${name}"
+
+maintainer="Erik Lundstedt <erik@lundstedt.it>"
+url="https://gitlab.com/Erik.Lundstedt/pacget.git"
+
+license="Don't Be a Jerk"
+depends="fzf curl bash"
+gives="$pkgname"
```
wizard-28, line 12:
```suggestion

```
Not needed!

packages/pacget-git/pacget-git.pacscript
```diff
@@ -0,0 +1,44 @@
+#!/bin/env bash
+#-*- mode: sh ;mode:lsp;-*-
+name="pacget-git"
+pkgname="pacget"
+pkgdir="/usr/src/pacstall/${name}"
+
+maintainer="Erik Lundstedt <erik@lundstedt.it>"
+url="https://gitlab.com/Erik.Lundstedt/pacget.git"
+
+license="Don't Be a Jerk"
```
wizard-28, line 10:
```suggestion
license="Don't Be a Jerk"
```
Not needed!

packages/pacget-git/pacget-git.pacscript
```diff
@@ -0,0 +1,44 @@
...
+depends="fzf curl bash"
+gives="$pkgname"
+description="A fully foss, lightweight fuzzysearching TUI frontend for pacstall using fzf"
+removescript="yes"
+
+debian=TRUE
+ubuntu=TRUE
```
wizard-28, line 17:
```suggestion
debian=TRUE
ubuntu=TRUE
```
Not needed!

packages/pacget-git/pacget-git.pacscript
```diff
@@ -0,0 +1,44 @@
...
+#this just coppies the files over and makes the file executable
+install() {
+	sudo install -Dm644 "LICENSE" "${pkgdir}/usr/share/licenses/${pkgname}/LICENSE"
+    sudo install -Dm755 "pacget" "/usr/src/$name/pacget/usr/bin/pacget"
+
+}
+
+removescript() {
+	rm -rf /usr/src/pacstall/pacget
+}
```
wizard-28, line 44:
```suggestion
removescript() {
	sudo rm -rf "/usr/share/licenses/${pkgname}"
}
```

No need to remove `${pkgdir}` (it gets deleted automatically), just delete the license directory

wizard-28 (changes requested):

packages/pacget-git/pacget-git.pacscript
erikLundstedt: i think its nice to have
although best case scenario would be if something similar got implemented directly into pacstall

wizard-28: okay.

packages/pacget-git/pacget-git.pacscript
erikLundstedt: booth the program itself and the pacscript have that licence and i prefer putting the licence-notice on there even if its not needed

packages/pacget-git/pacget-git.pacscript
```diff
@@ -0,0 +1,43 @@
...
+# pacget is a bashscript, no need to build anything
+build() {
+    true
+}
+#this just coppies the files over and makes the file executable
+install() {
+	sudo install -Dm644 "LICENSE" "${pkgdir}/usr/share/licenses/${pkgname}/LICENSE"
+    sudo install -Dm755 "pacget" "/usr/src/$name/pacget/usr/bin/pacget"
+
```
wizard-28, line 38:
```suggestion
	sudo install -Dm644 "LICENSE" "${pkgdir}/usr/share/licenses/${pkgname}/LICENSE"
	sudo install -Dm755 "pacget" "/usr/src/$name/pacget/usr/bin/pacget"
```

Fix spacing & remove empty line

wizard-28 (changes requested):

wizard-28 (approved):

PR #59, opened 2021-06-29
[Package Add] Tachidesk

D-Brox:
This PR adds `tachidesk`


packages/tachidesk/tachidesk.pacscript
```diff
@@ -0,0 +1,37 @@
+name="Tachidesk"
+pkgname="tachidesk"
+version="0.4.3_r670"
+description="A free and open source manga reader that runs extensions built for Tachiyomi"
+maintainer="D-Brox <BroxZuka@protonmail.com>"
+url="https://github.com/Suwayomi/Tachidesk.git"
+depends=("openjdk-15-jre")
```
wizard-28, line 7:
```suggestion
depends="openjdk-15-jre"
```
`depends` is not an array iirc.

wizard-28 (changes requested):
Is this a `-git` pacscript? You don't seem to be using any of the files cloned by git (except `LICENSE` etc), so why make it a `-git` pacscript?

D-Brox:
> Is this a `-git` pacscript? You don't seem to be using any of the files cloned by git (except `LICENSE` etc), so why make it a `-git` pacscript?

True, it doesn't really need to be a `-git`, I probably forgot to change it, since I just altered my `tachidesk-jui` script, which really needs to be a `-git` pacscript.

Should I just use `wget` to get the `README` and `LICENSE` files, and should I just remove the `.git` from the url?

wizard-28:
I don't think just downloading the `.jar` using `url` will work (check the CI) @Henryws needs to take a look here.

Henryws:
It won't work. It will try to untar it

D-Brox:
Speaking of which, I guess Pacstall should also deal with cases where there is no extension at the end of the url, since some packages can be bash scrips or executables (youtube-dl and czkawka, for example). I'll open an issue for feature reques later. For now I'll close this PR.

PR #64, opened 2021-07-03
[Package Add] surf-distrotube

WRM-42:


packages/surf-distrotube/surf-distrotube.pacscript
```diff
@@ -0,0 +1,34 @@
+name="surf-distrotube"
+pkgname="surf"
+pkgver() {
+    git ls-remote "${url}" master | cut -f1
+}
```
wizard-28, line 5: This needs to be written later as `url` isn't assigned yet, so it'll return nothing

wizard-28 (changes requested):

wizard-28 (approved):

PR #98, opened 2021-07-31
Feat: use newline instead of space in packagelist

rain1598:
Makes merging PRs a lot easier


Korkmatik:
I also think that would be much nicer if it works

PR #170, opened 2021-08-18
[Package Add] Brave-Browser-deb

WRM-42:


D-Brox (changes requested):
Needs to add the brave ppa to work, since it depends on brave-keyring

PR #264, opened 2021-09-10
Add thrive and fix packagelist

Henryws:


wizard-28 (changes requested):
Shouldn't it have a `gives="thrive-launcher"`

wizard-28 (approved):

packagelist
```diff
@@ -74,9 +88,9 @@ ms-teams-deb
 mullvad-vpn-deb
 mutt-wizard
 nano
-nemo
```
wizard-28, line 77: `nemo` not removed but not added again. Credits: @NarutoXY

wizard-28 (changes requested):

PR #277, opened 2021-09-14
fix(bpytop-git): Add `--upgrade` flag

wizard-28:
Fixes `bpytop-git`'s `psutil` installation. Previously if an older `psutil` version was installed then it was not updated. Which resulted in breakage. This fixes that.

Maintainer: @WRM-42


Henryws (approved):

WRM-42 (approved):
LGTM

PR #299, opened 2021-09-18
Add moka-icon-theme

aranym0:
Hello!
It's about time I contributed a new package..
Tested on Ubuntu VM, as usual


aranym0:
Installing this from the pacscript with pacstall -Il in the VM works perfectly. (I even just re-installed it to make sure)
Not sure what the failing check is about, I assume a problem in the environment.

Henryws:
Yeah, the CI has been broken for a while. The develop branch already has the fix in it, but we just need to release 1.6 and it should work again.

PR #304, opened 2021-09-19
add cantata-deb

ty1118:


packages/cantata-deb/cantata-deb.pacscript
```diff
@@ -0,0 +1,9 @@
+name="cantata-deb"
+pkgname="cantata-deb"
+version="2.4.2"
+url="https://launchpad.net/~ubuntuhandbook1/+archive/ubuntu/cantata-qt/+files/cantata_2.4.2-0build1~ubuntu20.10_amd64.deb"
```
wizard-28, line 4: Why is this pointing to `20.10`'s `.deb`. It won't work correctly in other versions of Ubuntu... I thought you would make a `nosuffix` ( no suffix at the end, builds the package first then installs ) package with the instructions given [here](https://github.com/CDrummond/cantata/blob/master/INSTALL).

See, if you don't want to make one yourself anymore you can always say so. Someone can volunteer to make one then.

And if your thinking making this `-deb` script was a waste of time. Then no, it wasn't it'll help the future volunteer (or you yourself! ) to make the correct script, as most of the tedious work like finding the dependencies etc is already done here.

wizard-28 (changes requested):

Henryws:
This PR was closed because:

1) your extracting the deb from a PPA. At that point, why not just use the PPA
2) You are using a version specific Deb file, so this will not work on newer versions of Ubuntu

ty1118:
Please do. I'm still learning. Baby steps are indeed needed.

PR #310, opened 2021-09-20
Adding in grapejuice script:

ty1118:
I have yet to test this script so I'm hoping that this works.


wizard-28 (changes requested):
Well... Can you delete and re-fork? As some unwanted changes are also getting added in this PR.

Before deleting copy the `grapejuie-git` pacscript somewhere else. Then paste it after re-fork.

Also, use a text editor and not the GitHub web editor.

wizard-28 (commented):
@ty1118 if you have the time could you join our [discord](https://discord.gg/yzrjXJV6K8)/[matrix](https://matrix.to/#/#pacstall:matrix.org)? So we could better communicate there?

PR #361, opened 2021-10-04
add tmux

harsh18262:
Add `tmux`

### DIFF
--- a/packages/tmux/tmux.pacscript
+++ b/packages/tmux/tmux.pacscript
@@ -1,0 +1,23 @@
+name="tmux"
+version="3.2a"
+url="https://github.com/tmux/tmux/releases/download/${version}/tmux-${version}.tar.gz"
+build_depends="gcc make pkg-config bison"
+depends="libncurses-dev libevent-dev"
+breaks="tmux-git"
+replace="tmux"
+description="A terminal multiplexer"
+hash="551553a4f82beaa8dadc9256800bcc284d7c000081e47aa6ecbb6ff36eacd05f"
+optdepends=("libutempter-dev: Record user sessions to utmp and wtmp files [add to depends array and rebuild to enable]")
+maintainer="Harshwardhan Mehrotra <abrakadabra18262@protonmail.com>"
+prepare() {
+          ./autogen.sh
+}
+build() {
+        ./configure
+        make -j$(nproc)
+}
+install() {
+          sudo make install DESTDIR="$STOWDIR/$name"
+          sudo install -Dm755 example_tmux.conf "$STOWDIR/$name/usr/share/tmux/example_tmux.conf"
+}
+
--- a/packages/tmux/tmux.pacscript
+++ b/packages/tmux/tmux.pacscript
@@ -3,21 +3,28 @@ version="3.2a"
 url="https://github.com/tmux/tmux/releases/download/${version}/tmux-${version}.tar.gz"
 build_depends="gcc make pkg-config bison"
 depends="libncurses-dev libevent-dev"
-breaks="tmux-git"
+breaks="${name}-git ${name}-bin ${name}-deb ${name}-app"
 replace="tmux"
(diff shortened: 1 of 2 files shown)


packages/tmux/tmux.pacscript
```diff
@@ -0,0 +1,23 @@
+name="tmux"
+version="3.2a"
+url="https://github.com/tmux/tmux/releases/download/${version}/tmux-${version}.tar.gz"
+build_depends="gcc make pkg-config bison"
+depends="libncurses-dev libevent-dev"
+breaks="tmux-git"
```
wizard-28, line 6:
```suggestion
breaks="${name}-git ${name}-bin ${name}-deb ${name}-app"
```

Also add the other package types.

packages/tmux/tmux.pacscript
```diff
@@ -0,0 +1,23 @@
...
+replace="tmux"
+description="A terminal multiplexer"
+hash="551553a4f82beaa8dadc9256800bcc284d7c000081e47aa6ecbb6ff36eacd05f"
+optdepends=("libutempter-dev: Record user sessions to utmp and wtmp files [add to depends array and rebuild to enable]")
```
wizard-28, line 10:
```suggestion
optdepends=("libutempter-dev: Record user sessions to utmp and wtmp files")
```

In version `1.6` (currently `develop`) the `optdepends` are called before the installation begins, just like `depends`, so no need to give this warning. The dependency will be available during install.

wizard-28 (changes requested):
https://github.com/archlinux/svntogit-community/blob/packages/tmux/trunk/PKGBUILD

---

**FROM THE AUR**

`install()`:
```
	install -D -m0644 COPYING "$pkgdir/usr/share/licenses/tmux/LICENSE"
```

---

Convert that line to the Pacstall compatible one, and add that.

packages/tmux/tmux.pacscript
```diff
@@ -0,0 +1,23 @@
...
+maintainer="Harshwardhan Mehrotra <abrakadabra18262@protonmail.com>"
+prepare() {
+          ./autogen.sh
+}
+build() {
+        ./configure
```
wizard-28, line 16:
```suggestion
        # Check if `libutempter-dev` ( optdepends ) is installed
        if dpkg-query -W libutempter-dev > /dev/null 2>&1; then
                ./configure --enable-utempter
        else
                ./configure
        fi
```

Suppose the user enables `libutempter`. You need to add the `--enable-utempter` flag here to enable it. My suggestion figures out if the user has installed the lib, and proceeds accordingly.

wizard-28 (changes requested):

harsh18262:
Suggested Changes added

wizard-28 (approved):
Well, you could have directly clicked the "Add suggestion to batch"/"Commit suggestion" button, but okay.

packages/tmux/tmux.pacscript
wizard-28: Fixed in https://github.com/pacstall/pacstall-programs/pull/361/commits/615b145c09a3c6bd509325f5cba8e977f1e09e50

packages/tmux/tmux.pacscript
wizard-28: Fixed in https://github.com/pacstall/pacstall-programs/pull/361/commits/615b145c09a3c6bd509325f5cba8e977f1e09e50

packages/tmux/tmux.pacscript
wizard-28: Fixed in https://github.com/pacstall/pacstall-programs/pull/361/commits/615b145c09a3c6bd509325f5cba8e977f1e09e50

harsh18262:
I had to add the command to install the license also so i thought it would create less commits.Will keep in mind for the next time
Thanks

PR #489, opened 2021-11-10
Remove Axel from CI

Henryws:
Axel causes the CI to fail downloading the source for a package.

I verified it is axels fault due to the line in the CI, where it say’s `error opening local file`
https://github.com/haileys/axel/blob/8e8eeb0650a097213d1b439e98629dfc6e709af9/axel.c#L178

### DIFF
--- a/.github/workflows/test-package-install.yml
+++ b/.github/workflows/test-package-install.yml
@@ -15,7 +15,7 @@ jobs:
         run: |
           export PACSTALL_SKIP_NETWORK_CHECK=1
           sudo bash -c "$(wget -q https://git.io/JsADh -O -)"
-          apt remove -y axel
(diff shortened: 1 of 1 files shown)


wizard-28:
> Axel causes the CI to fail downloading the source for a package.
> 
> I verified it is axels fault due to the line in the CI, where it say’s `error opening local file` https://github.com/haileys/axel/blob/8e8eeb0650a097213d1b439e98629dfc6e709af9/axel.c#L178

maybe the `SRCDIR` doesn't have the proper permissions. So it's unable to download the archive there?

Henryws:
We don't, we just keep on trying

D-Brox (approved):

wizard-28 (approved):
Let's hope for the best

PR #523, opened 2021-11-21
Add dino-git

Henryws:


wizard-28 (changes requested):
Add `gives` and expand `breaks`

---

Also install the license in the appropriate directory

wizard-28 (dismissed):
Add `dino-deb` to `breaks`

wizard-28 (approved):

PR #531, opened 2021-11-29
Add Bazecor-app

Henryws:


wizard-28 (changes requested):
Add `description`, `gives` and why is the post install script in separate file?

PR #535, opened 2021-11-30
fix yad

echometerain:
wtf

### DIFF
--- a/packages/yad/yad.pacscript
+++ b/packages/yad/yad.pacscript
@@ -32,7 +32,7 @@ prepare() {
 }
 
 build() {
-	./configure "$FLAGS"
+	./configure " $FLAGS"
 	make -j$(nproc)
 }
 


packages/yad/yad.pacscript
```diff
@@ -32,7 +32,7 @@ prepare() {
 }
 
 build() {
-	./configure "$FLAGS"
+	./configure " $FLAGS"
```
wizard-28, line 35: Umm... @echometerain , Can you explain why is this change necessary?

echometerain: It wouldn't build unless I do this

It likely resolved to `./configure--flag1 --flag2` instead of `./configure --flag1 --flag2`

wizard-28:
**DO NOT MERGE TILL #537 IS MERGED**

wizard-28 (changes requested):
Follow the instructions to setup your pacscript development environment listed [here](https://github.com/pacstall/pacstall-programs#how-to-setup-the-environment-for-pacscript-development)

After doing so try staging (`git add`) this file (make a small dummy modification) then try to commit it (after running `pre-commit install`). When the pre-commit hooks fail (if they fail), fix the errors.

echometerain:
`nothing to commit`

wizard-28 (approved):

PR #537, opened 2021-11-30
build: Add pre-commit hooks

wizard-28:
Fixes #536

PR #538, opened 2021-11-30
Add missing ubuntu dependencies

Henryws:
Taken from pipewires ubuntu CU


wizard-28:
**DO NOT MERGE TILL #537 IS MERGED**

packages/pipewire/pipewire.pacscript
```diff
@@ -15,6 +15,6 @@ build() {
 }
 
 install() {
-    cd builddir/
+    cd builddir/ || exit 1
```
wizard-28, line 18:
```suggestion
    cd builddir/ || return 1
```

If we use `return` then Pacstall can handle the error code properly instead of the program stopping execution at this point.

wizard-28 (changes requested):

wizard-28 (approved):

PR #545, opened 2021-12-07
Update make-a-pacscript to add `pkgname`

echometerain:


wizard-28 (commented):
Isn't it the same as `pkgname` though?

echometerain:
oh it's called pkgname

wizard-28 (approved):

PR #559, opened 2021-12-15
fix(fastfetch-git): update outdated pacscript

wizard-28:


packages/fastfetch-git/fastfetch-git.pacscript
```diff
@@ -66,10 +71,10 @@ removescript() {
 
   # Remove left over directories
   if ask "Do you want to remove the configuration files?" Y; then
-    sudo rm -rf "${homedir}/.config/fastfetch"
+    rm -rf "${homedir}/.config/fastfetch"
   fi
   if ask "Do you want to remove the left over cache?" Y; then
-    sudo rm -rf "${homedir}/.cache/fastfetch"
+    rm -rf "${homedir}/.cache/fastfetch"
```
Henryws, line 77: I would prefer the command `rm -rfI` instead of just `-rf`, because `I` will make it prompt the user for every removal after 3 files, which is way safer, especially when run in the home directory

Henryws (changes requested):

packages/fastfetch-git/fastfetch-git.pacscript
wizard-28: But right before I remove the files, I ask the user for explicit permission.

Henryws (approved):

PR #567, opened 2021-12-16
add: `oil`

wizard-28:
## Progress
- [x] Edit `packagelist`
- [x] Add initial pacscript
- [x] Contact devs: https://github.com/oilshell/oil/issues/1057
- [x] Add `maintainer` to pacscript: Resolved to `wizard-28` (https://github.com/oilshell/oil/issues/1057#issuecomment-997273484)


packages/osh/osh.pacscript
```diff
@@ -0,0 +1,30 @@
+name="osh"
+pkgdir="${STOWDIR}/${name}"
+version="0.9.5"
+description="Oil Shell - A bash-compatible UNIX shell written in Python"
```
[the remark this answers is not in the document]
wizard-28: Actually `osh` also installs `oil`. This package installs 2 things: `oil` & `osh`

packages/osh/osh.pacscript
```diff
@@ -0,0 +1,37 @@
...
+# Edit MAKEFLAGS here
+MAKEFLAGS=
+
+prepare() {
+  mkdir -p tmp
+
+  # HACK: Change `TMP` directory to `./tmp/` for building
+  # We do not want this to expand
+  # shellcheck disable=SC2016
+  sed 's/TMP=${TMPDIR:-\/tmp}/TMP=.\/tmp/' -i "configure"
```
andychu, line 19: TMPDIR=./tmp/ is a lot more reliable than this

packages/osh/osh.pacscript
```diff
@@ -0,0 +1,37 @@
...
+
+  # HACK: Change `DESTDIR` by inserting it directly into the file
+  #       Since `DESTDIR="${pkgdir}/ sudo ./install" doesn't work
```
andychu, line 22: Try  `sudo DESTDIR=x ./install`

One way to debug the difference is trying:

    sudo DESTDIR=x env

vs

    DESTDIR=x sudo env

andychu (commented):
Thanks for packaging Oil.  I think it's better to call it `oil` as mentioned here.

https://github.com/oilshell/oil/wiki/Oil-Deployments (please update the wiki when/if this is merged)

I made comments on 2 shell issues that are unnecessarily fragile.

packages/osh/osh.pacscript
wizard-28: Well, that works but when testing I found out that the `Makefile` makes a directory called `_tmp`, where apparently no file is stored throughout the whole build. So, is it okay to point `TMP` to that directory?

packages/osh/osh.pacscript
wizard-28: Works! I'll push after https://github.com/pacstall/pacstall-programs/pull/567#discussion_r771897111 gets resolved.

packages/osh/osh.pacscript
andychu: I'm just saying you shouldn't use sed to change the value of the variable.  You should just use `TMPDIR=whatever ./configure`.

This kind of sed usage is bad because it can easily break later, and is unnecessary

Henryws (approved):

PR #642, opened 2022-01-09
fix(neovim-git): modernize

Henryws:


wizard-28 (changes requested):
NVM

wizard-28 (changes requested):
Add `gives`, `replace`, `breaks`

wizard-28 (approved):
Approving, as it works locally as stated by @Henryws

PR #643, opened 2022-01-11
fix(walc-app): modernize

wizard-28:


Henryws (changes requested):
Add `pkgname`

wizard-28:
> Add `pkgname`

Not required

PR #658, opened 2022-01-17
upd(spotify-client-deb): `1.1.68.632.g2b11de83` ->  `1.1.72.439.gc253025e`

smokeythemonkey:
updated spotify version and has as the old version is no longer available at that url


Henryws (approved):

packages/spotify-client-deb/spotify-client-deb.pacscript
```diff
@@ -2,8 +2,8 @@ name='spotify-client-deb'
 gives='spotify'
```
Henryws, line 2:
```suggestion
gives='spotify-client'
```

Henryws (changes requested):

Henryws (approved):

PR #663, opened 2022-01-23
rem: `makedeb-makepkg-deb`

cat-master21:
Makedeb-makepkg has disappeared from Makedeb repo, AUR, and from Proget meaning that this pacscript will not work anymore.


wizard-28 (commented):
This will break [makedeb-deb](https://github.com/pacstall/pacstall-programs/blob/fab548433e650d944e3c41778d00116d60b42140/packages/makedeb-deb/makedeb-deb.pacscript) too, as it relies on `makepkg`

cat-master21:
Updated Makdeb-deb to newer version which doesn't need Makepkg-deb but conflicts with it. Should be ready now.

wizard-28 (approved):
Looks good to me, but requesting a review from @Henryws

Henryws (approved):

PR #672, opened 2022-01-26
upd(android-studio): `2020.3.1.25` -> `2021.1.1.20`

cat-master21:


Henryws (approved):

Henryws:
Also as a heads up, use conventional commits for commit messages

---

conventionalcommits.org/

cat-master21:
Issues with installing so plan to fix.

wizard-28:
Is this ready for review? Why it a draft still?

cat-master21:
Android-studio does not install properly and if I were to update GitHub there would be a lot of commits. The current state on my end is it can't find $STUDIO_JDK

packages/android-studio/android-studio.pacscript
```diff
@@ -2,33 +2,33 @@ name="android-studio"
...
 install() {
-    # Moving all android studio files to /opt
-    sudo cp -a bin lib jre plugins license LICENSE.txt build.txt product-info.json "$STOWDIR/$name/opt/$name"
+    # Moving android studio files to /opt
+    sudo cp -a bin lib jre plugins license LICENSE.txt build.txt product-info.json NOTICE.txt brokenPlugins.db icons.db classpath.txt "$STOWDIR/$name/opt/$name"
 
     # Creating a launcher in $PATH
-    sudo ln -s "$STOWDIR/$name/opt/$name/studio.sh" "$STOWDIR/$name/usr/bin/android-studio"
-
+    sudo ln -sf "$STOWDIR/$name/opt/$name/bin/studio.sh" "/usr/bin/android-studio"
```
Henryws, line 20:
```suggestion
    sudo ln -sf "$STOWDIR/$name/opt/$name/bin/studio.sh" "$STOWDIR/$name/usr/bin/android-studio"
```

Henryws (changes requested):

packages/android-studio/android-studio.pacscript
cat-master21: I tried this but I get the error: source is an absolute symlink, and fails to install. No matter what I try I get this error except for a hard link, but that doesn't point to JDK or other android-studio files.

Henryws: How about moving `studio.sh` to `android-studio`?

cat-master21: Same thing as hard-linking; can't find JDK or other android-studio files.

cat-master21:
Still once more, same issue, Android-studio can't find JDK and other needed files.
Screenshot
:

cat-master21:
Because there doesn't appear to be a solution linking files this PR will be closed.

PR #673, opened 2022-01-27
fix(nano):  make and automake edge case

D-Brox:
Replace `make` position to avoid wrong first match substitution (`automake -> auto`)


D-Brox:
@cat-master21 This should fix the issue for now

Henryws (approved):

wizard-28 (approved):

PR #717, opened 2022-02-09
upd(dotdrop): `1.7.1` -> `1.8.2`

cat-master21:


wizard-28 (approved):
TZ data problem with some OSes in the CI but LGTM.

PR #793, opened 2022-02-28
add `emby-server-deb`

warofzen:
Add a pacscript to install [Emby Server](https://github.com/MediaBrowser/Emby.Releases) from their .deb release


packages/emby-server-deb/emby-server-deb.pacscript
```diff
@@ -0,0 +1,8 @@
+name="emby-server-deb"
+version="4.6.7.0"
+gives="emby-server"
+breaks="${gives} ${gives}-bin ${gives}-app"
```
wizard-28, line 4:
```suggestion
breaks="${gives} ${gives}-bin ${gives}-app ${gives}-git"
```

wizard-28 (changes requested):
@Henryws I think it's okay to not contact the devs for this.

wizard-28 (approved):

wizard-28:
@warofzen the CI fails in all OSes

warofzen:
It seems that the CI cannot create a symlink for the `emby-server.service` file, but it worked just fine on my local machine:
https://asciinema.org/a/DwwakO0Zdpf6LiibqCtP1WniB

wizard-28:
> It seems that the CI cannot create a symlink for the `emby-server.service` file, but it worked just fine on my local machine: https://asciinema.org/a/DwwakO0Zdpf6LiibqCtP1WniB

Okay, so @Henryws will test it on his machine, and then we can merge.

Issue #810, opened 2022-03-05
tmux.pacscript: line 13: ./autogen.sh: No such file or directory


HarleyQuinn-Z6:
```
$ pacstall -I tmux
Do you want to view/edit the pacscript [y/N]
[+] INFO: Sourcing pacscript
[+] INFO: URL exists
[+] INFO: Creating dummy package
[sudo] password for user:
[+] INFO: tmux has optional dependencies that can enhance its functionalities
Optional dependencies:
    libutempter-dev: Record user sessions to utmp and wtmp files
Do you want to install them [Y/n]
Reading package lists... Done
Building dependency tree
Reading state information... Done
Note, selecting 'tmux' instead of '/tmp/pacstall/tmux-pacstall.deb'
Suggested packages:
  libutempter-de
The following NEW packages will be installed:
  tmux
0 upgraded, 1 newly installed, 0 to remove and 7 not upgraded.
After this operation, 0 B of additional disk space will be used.
Get:1 /tmp/pacstall/tmux-pacstall.deb tmux all 3.2a-1 [1260 B]
Selecting previously unselected package tmux.
(Reading database ... 328823 files and directories currently installed.)
Preparing to unpack .../tmp/pacstall/tmux-pacstall.deb ...
Unpacking tmux (3.2a-1) ...
Setting up tmux (3.2a-1) ...
[+] INFO: Installing dependencies
[+] INFO: Retrieving packages
Initializing download: https://github.com/tmux/tmux/releases/download/3.2a/tmux-3.2a.tar.gz
File size: 648394 bytes
Opening output file tmux-3.2a.tar.gz
Starting download

Connection 0 finished
Connection 1 finished
Connection 2 finished
Connection 3 finished
Connection 0 unexpectedly closed
Connection 1 finished

Downloaded 633.197 Kilobyte(s) in 0 second(s). (806.50 KB/s)
[+] INFO: Preparing
tmux.pacscript: line 13: ./autogen.sh: No such file or directory
[!] ERROR: Could not prepare tmux properly
[+] INFO: Cleaning up
```
 - OS: Linux Mint
 - Version 20.3

wizard-28:
The maintainer is @harsh18262

PR #528, opened 2021-11-25
feat(packagelist): Add `spacecadetpinball-git`

wizard-28:
Closes #472

PR #811, opened 2022-03-05
fix: tmux pacscript 

harsh18262:
Fix for Issue #810


wizard-28 (changes requested):
Check the CI, the build fails.

harsh18262:
That is the problem from ubuntu 20.04 as stow is not available in ubuntu 20.04 repositories we will need to add bionic repos and then install stow after that it will install success fully

harsh18262:
I think there is a problem with pacstall it is not installing optional dependencies on ubuntu ,it is installing on debian .I also tried on ubuntu bionic docker image with the neofetch package for that package also pacstall is not installing optional dependencies

wizard-28:
@Henryws can you reproduce this?

Henryws:
> @Henryws can you reproduce this?

Yes (we should probably make apt install suggested packages), and also, the optdepends package should be in the depends variable, as it fails to build without it.

wizard-28:
Linking: https://github.com/pacstall/pacstall/pull/519

Henryws (approved):

wizard-28 (changes requested):
> `configure: error: "utempter not found"`

Check the CI logs, and add that dependency.

harsh18262:
> > `configure: error: "utempter not found"`
> 
> Check the CI logs, and add that dependency.

It will be fixed as soon as PR [#528](https://github.com/pacstall/pacstall/pull/528) in pacstall repo is merged. The issue is that the optional dependencies is not being installed

PR #846, opened 2022-03-15
feat(soundux): add repology variable

D3SOX:
As described in #845


wizard-28 (approved):

wizard-28:
@D3SOX, The build fails on Debian unstable as `libappindicator3-dev` is no longer in their repositories. Merging anyways.

PR #857, opened 2022-03-19
upd(librewolf-app): `97.0.1-1` -> `98.0`

wizard-28:
Also add repology


wizard-28:
@Henryws timeout merging.

PR #861, opened 2022-03-19
fix(emacs-git): fix everything

wizard-28:


wizard-28:
@Henryws timeout. Merging.

PR #885, opened 2022-03-26
add: `piavpn`

WRM-42:


wizard-28 (changes requested):
Dismiss this review, requested changes by accident.

packages/piavpn/piavpn.pacscript
```diff
@@ -0,0 +1,22 @@
+name="piavpn"
+breaks="${name}-git"
+repology=("project: ${name}")
```
wizard-28, line 3: Won't work as the package in repology uses a different format of the same version, IDK if @Henryws could device a method to automatically convert the repology version format to a usable one for this pacscript.

wizard-28 (changes requested):

packages/piavpn/piavpn.pacscript
Henryws: I could, but I don't know how that would work in the context of the repology variable

wizard-28: > url="https://installers.privateinternetaccess.com/download/pia-linux-${version}.run"

As you can see the current `version` variable is being directly embedded in the `url` link. So I need you to device a method to automatically convert the version in the repology screenshot to a format that can be used by this link.

I mean just assume the `version` in the pacscript is the repology version, and then write code to make that `version` variable usable in the `url`.

packages/piavpn/piavpn.pacscript
```diff
@@ -0,0 +1,22 @@
...
+prepare() {
+    true
+}
+
+build() {
+    true
+}
+
+install() {
+    chmod +x pia-linux-${version}.run
```
Henryws, line 19: Nothing gets installed to `/usr/src/pacstall`, which results in a stow error, which breaks the build

Henryws (changes requested):

packages/piavpn/piavpn.pacscript
Mar0xy: Yea I think this should actually follow the PKGBUILD of piavpn-bin properly so it works correctly with pacstall since the AUR package extracts the files from the run file and then installs them manually by copying the files to the locations.

packages/piavpn/piavpn.pacscript
```diff
@@ -0,0 +1,47 @@
...
+
+build() {
+  true
+}
+
+install() {
+  chmod +x pia-linux-${version}.run
+  ./pia-linux-${version}.run --noexec --target "${SRCDIR}/${name}"
+  sudo install -Dm755 "piafiles" "${STOWDIR}/${name}"
+  sudo install -Dm755 "installfiles" "${STOWDIR}/${name}/bin"
```
Mar0xy, line 23: you forgot to cd into the directory it extracted the files into also these install lines don't work if you want to copy folders and rename them you have to use cp.

Mar0xy (changes requested):

PR #892, opened 2022-03-27
fix(yad): fix hash

cat-master21:


cat-master21:
Some issues with yad dependencies also after some testing.

wizard-28:
LGTM, but requesting a review from @Henryws

Henryws (approved):
👍🏻

PR #893, opened 2022-03-27
fix(android-studio): installation

raxod502:
Running `pacstall -I android-studio` previously failed with the following error for me:

```
[+] INFO: Preparing
[+] INFO: Building
[+] INFO: Installing
cp: target '/usr/src/pacstall/android-studio/opt/android-studio' is not a directory
ln: failed to create symbolic link '/usr/src/pacstall/android-studio/usr/bin/android-studio': No such file or directory
/usr/src/pacstall/android-studio/usr/share/applications/android-studio.desktop: No such file or directory
chmod: cannot access '/usr/src/pacstall/android-studio/opt': No such file or directory
[!] ERROR: Could not install android-studio properly
[+] INFO: Cleaning up
```

If I understand correctly, this is because Pacstall does not pre-emptively create any directories within the package installation tree at `/usr/src/pacstall/$name`. This pull request uses `install -d` to create the needed parent directories.

Once that problem was addressed, I got the following error from Stow:

```
[+] INFO: Symlinking files
WARNING! stowing android-studio would cause conflicts:
  * source is an absolute symlink ../../usr/src/pacstall/android-studio/usr/bin/android-studio => /usr/src/pacstall/android-studio/opt/android-studio/bin/studio.sh
All operations aborted.
[!] ERROR: Package contains links to files that exist on the system
[+] INFO: Cleaning up
[!] ERROR: Failed to install android-studio
```

If I understand correctly, this is because files in the source directory should be referencing paths directly on the target filesystem, and not paths in the source directory itself. Adjusting the link target from `$STOWDIR/$name/usr/bin/android-studio` to `/usr/bin/android-studio` fixed the error.

With this PR, I can successfully install and run Android Studio via Pacstall. However, I may be missing something, because otherwise I don't understand how the package worked correctly for anyone else, including the initial contributor.


cat-master21:
Yeah I tried that to: https://github.com/pacstall/pacstall-programs/pull/672 But it failed with JDK environment variable.

raxod502:
I don't know offhand what might cause that on your system, but the Pacscript as updated in this PR works for me on Pop!_OS 21.10 with OpenJDK 11.0.14 and other packages updated as of 2022-03-27, and Android Studio boots fine.

raxod502:
CI appears to be red on master with the same errors as on this PR, so I don't believe any new problems were introduced by the change.

cat-master21:
Yeah, it works because you linked it directly with `sudo ln -s "$STOWDIR/$name/opt/android-studio/bin/studio.sh" /usr/bin/android-studio` but @Henryws doesn't like that: https://github.com/pacstall/pacstall-programs/pull/672#discussion_r795311352.

raxod502:
@Henryws Mind adding some context about why having `/usr/bin/android-studio` as a symlink would be an issue? Collapsing the symlink has the potential to cause problems by changing the runtime environment slightly - for example, if the script tries to retrieve its own filename to determine the installation directory, that'll break unless it is executed through a symlink.

raxod502:
Oh, gosh, I made a terrible mistake here. What I have is wrong; the first argument is the one that needs to be absolute, not the second. (I'm glad we're planning to move away from using sudo in these build scripts; it can lead to very dangerous missteps like this one.)

Henryws:
It's primarily an issue of the package all being in one place, that being `$STOWDIR/$name`. Having to symlink a file from an already symlinked directory seems very hacky.

raxod502:
I see, the issue is that Stow can't handle installing packages that contain symlinks to absolute files. I'm not sure I understand the reason for that limitation, as I can't think of any technical issue - this sort of thing happens all the time with distro packages and is supported by all Linux package managers that I know of - but in any case it's possible to work around the issue by using a relative symlink instead. With the latest commit, Android Studio is again installable and boots fine on my system.

raxod502:
CI is fixed by https://github.com/pacstall/pacstall-programs/pull/894.

Henryws (approved):

raxod502:
Are further changes required for this PR to be mergeable?

Henryws:
> Are further changes required for this PR to be mergeable?

Nothing from me; this PR works perfectly in a test docker container.

Henryws:
All the tests pass, and the PR works, so merging.

PR #894, opened 2022-03-27
Fix CI

raxod502:
I notice CI has been red on master for some time; this PR should fix that. Marking it as a draft until CI is green.

PR #910, opened 2022-04-02
ci(fix): Assign PR Author as Assignee

Mar0xy:
I have noticed that it keeps failing to Assign the PR author and replaced it with a more up to date PR assign action while also trying it out on the fork.


wizard-28:
Already being done by @raxod502 in https://github.com/pacstall/pacstall-programs/pull/894

PR #911, opened 2022-04-02
upd(opera-deb): `85.0.4341.18` -> `85.0.4341.47`

wizard-28:


Henryws (approved):
Opera has a stupid dpkg hook that asks if the user wants to add the opera repo to the system, but the CI got this far, so I assume it works

PR #925, opened 2022-04-03
upd(makedeb-deb): `10.0.0` -> `11.0.1-1`

cat-master21:


packages/makedeb-deb/makedeb-deb.pacscript
```diff
@@ -1,6 +1,7 @@
 name="makedeb-deb"
-version="10.0.0"
-url="https://proget.hunterwittenborn.com/debian-feeds/makedeb/main/makedeb/all/makedeb_all_${version}-1.deb"
+version="11.0.1-1"
```
Henryws, line 2: Keep the version without the `-1` suffix.

Henryws (changes requested):

wizard-28 (approved):

packages/makedeb-deb/makedeb-deb.pacscript
```diff
@@ -1,6 +1,7 @@
 name="makedeb-deb"
-version="10.0.0"
+version="11.0.1"
+repology=("project: makedeb")
 url="https://proget.hunterwittenborn.com/debian-feeds/makedeb/main/makedeb/all/makedeb_all_${version}-1.deb"
```
wizard-28, line 4: Hmm, Pacup might not work due to these suffixes being present. Say in the next version they named the file as `makedeb_all_11.0.2.deb`, as you can see the version suffix isn't present anymore. So when pacup tries to replace `${version}` with the new version from repology (`11.0.2`), then the resulting file name would be `makedeb_all_11.0.2-1.deb`.

I don't think anything can be done about this at the moment. Still keep the `repology` variable for stats in our website.

PR #1017, opened 2022-04-22
Shorten hashes for -git packages

Mar0xy:
Repology when


Mar0xy:
Well that worked better than expected :rofl:

Mar0xy:
image
image

Henryws (changes requested):
I believe you can use cut to extract a substring. Don't use awk, it's too big for a simple extraction

Mar0xy:
> I believe you can use cut to extract a substring. Don't use awk, it's too big for a simple extraction

You can but then you would need to set another pipe just to substring it via cut since you can't do multiple arguments in the same command

Mar0xy:
There that is how the situation would be when using cut.

Mar0xy:
@Henryws If you are fine with this solution of using cut I could change all git packages which currently use this method in a PR to make them use 8 char long hashes.

Henryws:
Yeah using two instances of `cut` will be faster than calling `awk`

Henryws (approved):

PR #1064, opened 2022-04-29
upd(epson-printer-utility-deb): added repology

Zahrun:


Henryws (changes requested):
CI fails

Zahrun:
> CI fails

As soon as debian unstable develop fails, the other jobs stopped `The job was canceled because "debian_unstable_develop" failed.`
This pacscript has dependencies that I was only able to find in a ppa, hence it will not work on debian.

Henryws:
What to do @wizard-28

cat-master21:
The Debian alt is https://packages.debian.org/sid/lsb-compat though amd64 version is for stretch only. Debian has indeed removed it though except for base packages https://wiki.debian.org/DebianLsb

wizard-28:
> What to do @wizard-28

So is there any way to fix the issue on Debian? Does it work with the base package in Debian as @cat-master21 pointed out?

If not, then disable it's installation on Debian and call it a day.

cat-master21:
The debian alternative is `lsb-compat` https://packages.debian.org/stretch/amd64/lsb-compat/filelist and https://packages.ubuntu.com/jammy/amd64/lsb-core/filelist note that lsb is just a meta package that installs this and also `lsb-printing` https://packages.ubuntu.com/jammy/lsb-printing which is yet another meta package. It is probably one of those dependencies that this deb actually needs. Also see https://wiki.debian.org/DebianLsb for more info.

PR #1094, opened 2022-05-07
upd(nala-deb): `0.8.1` -> `0.8.2`

volitank:


cat-master21:
Use [libpython3-dev](https://packages.ubuntu.com/jammy/libpython3-dev) which after looking in end rely's on latest libpython version @volitank

cat-master21:
Use [libpython3-dev](https://packages.ubuntu.com/jammy/libpython3-dev) which after looking in end rely's on latest libpython version @volitank

volitank:
> Use [libpython3-dev](https://packages.ubuntu.com/jammy/libpython3-dev) which after looking in end rely's on latest libpython version @volitank but it may be to much extra stuff though

Won't work on libpython3.10 I'm pretty sure. Either way doesn't really solve anything as on newer systems it should be installing Nala not Nala legacy.

I think the proper fix to this is flipping my if statement to check for which systems should run legacy instead of the way it is now

cat-master21:
Well if your doing the if statement way, you could create a pacscript with a if statement on pacdeps, to install either nala-legacy-deb, or nala-new-deb and pacscript would be named nala-deb so a smooth transition if nala-deb ups it's python version for example then the user updates and it switches to nala-legacy-deb for example.

Henryws (approved):

PR #1114, opened 2022-05-10
add: `surfshark-vpn-cli-deb`

WRM-42:


Henryws (changes requested):
Add `breaks`

WRM-42:
Quite redundant since Surfshark only provides deb packages

Henryws (approved):

PR #1135, opened 2022-05-15
add: `signal-desktop-deb`

jonafato:
- [x] Edit packagelist
- [x] Add initial pacscript
- [ ] Contact devs


packages/signal-desktop-deb/signal-desktop-deb.pacscript
```diff
@@ -0,0 +1,10 @@
+name="signal-desktop-deb"
+pkgname="signal-desktop"
+gives="signal-desktop"
+replace="signal-desktop"
```
Henryws, line 4:
```suggestion
gives="${pkgname}"
replace="${pkgname}"
```

Henryws (changes requested):

packages/signal-desktop-deb/signal-desktop-deb.pacscript
```diff
@@ -0,0 +1,10 @@
+name="signal-desktop-deb"
+pkgname="signal-desktop"
+gives="${pkgname}"
+replace="${pkgname}"
+repology=("project: signal-desktop")
```
WRM-42, line 5:
```suggestion
repology=("project: ${pkgname}")
```

WRM-42 (changes requested):
Add `breaks`

jonafato:
@WRM-42 Thanks for the review.

> Add `breaks`

Should this also be `${pkgname}`? This is my first contribution here, so I'm still learning the project conventions.

WRM-42:
> Should this also be `${pkgname}`? This is my first contribution here, so I'm still learning the project conventions.

You should also add `${pkgname}`-beta and `${pkgname}`-app, since Signal has these two versions

jonafato:
> > Should this also be `${pkgname}`? This is my first contribution here, so I'm still learning the project conventions.
> 
> You should also add `${pkgname}`-beta and `${pkgname}`-app, since Signal has these two versions

Per the note copied below from [the beta support article](https://support.signal.org/hc/en-us/articles/360007318471-Signal-Beta), the beta is a separate install from the regular desktop app and explicitly does not conflict.

>This is a new install with no message history. It is separate from the [production version of Signal Desktop](https://signal.org/download) and will not replace it.

I'm happy to add the `-app` version here, though I'm not sure I see where Signal provides an appimage build (or am I misunderstanding the meaning of `-app` in this context?).

(I'm planning on packaging the beta as well once this addition is in a good merge-ready condition.)

Henryws:
What `${pkgname}-app` means, when put in `breaks` is that *if* an appimage was released, this pacsctipt would be able to detect that.

WRM-42:
>Per the note copied below from [the beta support article](https://support.signal.org/hc/en-us/articles/360007318471-Signal-Beta), the beta is a separate install from the regular desktop app and explicitly does not conflict.

Thanks for mentioning this

> I'm happy to add the `-app` version here, though I'm not sure I see where Signal provides an appimage build (or am I misunderstanding the meaning of `-app` in this context?).

I've seen some people using AppImages of Signal, but they're not official

WRM-42 (approved):
++

PR #1145, opened 2022-05-21
upd(bitwig-studio-deb): `4.2` -> `4.2.3`

Zahrun:


Zahrun:
I guess we need
```
sudo dpkg --add-architecture i386
sudo apt-get update
```
Should it be in `prepare`?

Henryws:
> I guess we need
> 
> ```
> sudo dpkg --add-architecture i386
> sudo apt-get update
> ```
> 
> Should it be in `prepare`?

This should be in the CI actually, we'll fix that soon.

Henryws (approved):
Works in docker.

PR #1166, opened 2022-05-23
upd(qogir-gtk-theme-bin): `2021-12-25` -> `2022-04-29`

wizard-28:


Henryws (changes requested):
Fix dependencies

cat-master21:
This package is in Ubuntu (https://packages.ubuntu.com/kinetic/gtk3-engines-unico). But the debian alternative is missing? (https://packages.debian.org/search?suite=sid&arch=any&mode=filename&searchon=contents&keywords=libunico.so). See https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=822976 and also https://tracker.debian.org/news/791081/removed-103bzr152-1-from-unstable/

cat-master21:
Also I do not think that this needs gtk3-engines-unico anyway according to the README nothing is mentioned. This includes the [arc-theme](https://github.com/horst3180/Arc-theme) which the theme is based off.

PR #1195, opened 2022-05-29
add: `qtile-extras-git`

Ruturajn:
- Added `qtile-extras-git` to the list of supported packages.
- Added `qtile-extras-git.pacscript`.
- Updated the `packagelist`.

The package `qtile-extras-git` installs `qtile-git` as a dependency, but I am not sure as to, how to uninstall `qtile-git`, when `$ pacstall -R qtile-extras-git` is called. I was thinking that running `pacstall -R qtile-git` in the `removescript` function, would do the job, is this the right way, if not please let me know how I should do this.

Thanks

### DIFF
--- a/packagelist
+++ b/packagelist
@@ -209,6 +209,7 @@ protonvpn-nm-lib
 pycharm-community-bin
 qogir-gtk-theme-bin
 qogir-icon-theme-bin
+qtile-extras-git
 qtile-git
 realvnc-vnc-viewer-deb
 renegate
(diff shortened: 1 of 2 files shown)


Henryws (approved):

Ruturajn:
Is it fine, if `qtile-git` is not removed when `pacstall -R qtile-extras-git` is called? Because I hadn't put it in the pacscript.

Henryws:
`qtile-extras-git` is a supplementary package to `qtile-git`, so if it gets removed, it shouldn't remove `qtile-git`.

Ruturajn:
Ok, great 👍

PR #1196, opened 2022-05-29
fix(github-desktop-deb): repology version fix

Henryws:


cat-master21 (changes requested):
Have it where there is a separate variable for the full version like the PKGBUILD in https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=github-desktop

Henryws:
I don't think we need a separate variable for the `-linux2`

cat-master21:
Well we definitely don't need one but it may be better to have a variable. I'm fine if don't though.

Henryws:
Well, it seems that they change the suffix every other release, so it might be useful to add a variable. For now, I think not adding one is fine.

cat-master21 (approved):

Issue #1256, opened 2022-06-06
PacReq: `authy`


saenai255:
### Types

-deb (Debian achieve release of the package)

### URL

https://snapcraft.io/authy

### Building guide

Extract it from the snap package

PR #1259, opened 2022-06-07
add: `authy-bin`

saenai255:
Closes https://github.com/pacstall/pacstall-programs/issues/1256


Henryws (changes requested):
Mixed use of quotes (" and ') in variables

PR #1301, opened 2022-06-14
upd(bottles): `2022.5.28-trento-3` -> `2022.6.14-brescia`

cat-master21:
Bottles is now gtk4 + libadwaita! So the dependencies will probably be updated.

### DIFF
--- a/packages/bottles/bottles.pacscript
+++ b/packages/bottles/bottles.pacscript
@@ -1,9 +1,9 @@
 name="bottles"
-version="2022.5.28-trento-3"
+version="2022.6.14-brescia"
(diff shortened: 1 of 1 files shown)


cat-master21:
Dropping this pacscript may be done cause of https://usebottles.com/blog/an-open-letter/ @Mar0xy did you see this?

Mar0xy:
I think it's wrong to force Flatpak down people's throat and way to biased considering even ubuntu doesn't support flatpak out of the box but snap instead i myself use arch and also don't agree with them planning to remove the AUR package as I don't think Flatpak is the best solution.

Henryws (approved):

PR #1352, opened 2022-07-05
upd(windscribe-gui-deb): `2.3.15_beta` -> `2.4.8_beta`

WRM-42:


Henryws (approved):
Works, seems that docker doesn't contain systemctl

PR #1357, opened 2022-07-06
add: `github-cli-git`

WRM-42:


Henryws:
Add `go` to deps list

WRM-42:
It's already there:

`pacdeps=("go-bin")
`

Henryws:
Seems to be a bug with pacdeps in the develop branch, gonna manually test on master.

Henryws:
image
@saenai255 issue with `go-bin`

Henryws:
@WRM-42 Can you test by running the following commands, then screenshotting:
```bash
pacstall -U Henryws fix-pacdeps-again
pacstall -I $your_scripts
# screenshot them, upload them here
pacstall -U pacstall master # if you want to go back to the master branch
```

WRM-42:
Pacdep now works correctly, looks like an issue with Go
ubuntu_HclNvkywCW
ubuntu_Ddqqy5ft4E

Mar0xy:
I guess the go issue has to do with the fact that the root has to be changed as an env variable which is done by the pacstall script but you also need to source/update the current environment to have the change be applied which doesn't obviously happen in this case causing it to break.

Henryws:
So what should we do?

cat-master21:
Maybe perhaps instead use distribution's golang-go: https://packages.ubuntu.com/jammy/golang-go

PR #1369, opened 2022-07-10
fix(discord): cleaner commands

Henryws:


packages/discord/discord.pacscript
```diff
@@ -17,7 +17,7 @@ build() {
 }
 
 install() {
-        sudo mv discord.desktop /usr/src/pacstall/discord/usr/share/applications/
-        sudo cp discord.png /usr/src/pacstall/discord/usr/share/icons/
-        sudo mv ./* /usr/src/pacstall/discord/usr/share/discord/
+        sudo install -Dm755 discord.desktop "${STOWDIR}/${name}/usr/share/applications/discord.desktop"
+        sudo install -Dm755 discord.png "${STOWDIR}/${name}/usr/share/icons/discord.png"
+        sudo cp -R ./* "${STOWDIR}/${name}/usr/share/discord/"
```
wizard-28, line 22: Why is this a copy? Won't this duplicate the files temporarily?

Henryws: The previous command `mv` would fail if `$STOWDIR/$name/usr/share/discord` had files in it, meaning this would fail on upgrades

PR #1372, opened 2022-07-11
fix(bitwarden-deb): change repology

mirenradia:
It seems that the [bitwarden-desktop repology entry](https://repology.org/project/bitwarden-desktop/versions) is not updated as often as the [bitwarden repology entry](https://repology.org/project/bitwarden/versions) so I have changed the `bitwarden-deb` package to use the latter. Note that this pacstall package appears under the new repology entry (and *not* the old one). I have also checked a few other package repositories to confirm that their packages are for the desktop app (as opposed to the CLI tool or something else related to bitwarden).


mirenradia:
Any idea why the tests failed? I haven't changed anything other than the repology :thinking:

mirenradia:
You were right. After rebasing onto master and pushing again, it seems to be passing.

Henryws (approved):

PR #1387, opened 2022-07-13
upd(vscode-deb): `1.69.0` -> `1.69.1`

charitarthchugh:


Henryws:
For future reference, removing the `v` suffix from versions and use the character "`" around the versions.

Henryws (approved):

PR #1442, opened 2022-07-30
upd(git): `2.36.1` -> `2.37.1`

cat-master21:


Henryws (approved):
Works in a VM when I remove `git` from apt

PR #1454, opened 2022-08-05
Added systemcoding-conky

systemcoding:


systemcoding:
Didn't add packagelist

PR #1480, opened 2022-08-13
add: `nerd-fonts-jetbrains-mono`

dbednar230:
Same layout as `nerd-fonts-hack`, with different font.


dbednar230:
Commit was given the wrong name - missing an "s" in the "fonts", will fix

dbednar230:
Just going to make a new PR because I messed up like 3 times

PR #1526, opened 2022-08-28
add: `wsysmon-bin`

0x262D:


packages/wsysmon-bin/wsysmon-bin.pacscript
```diff
@@ -0,0 +1,30 @@
+maintainer="0x262D <pacstall@copypaste.anonaddy.me>"
+
```
Henryws, line 2:
```suggestion
```

Henryws (changes requested):
Does this program have an icon?

0x262D:
> Does this program have an icon?

This project does not have an icon, but I found a similar project with an icon that could be used. Is it worth it to do so?

Icon
https://github.com/KrispyCamel4u/SysMonTask/blob/master/icons/SysMonTask.png

packages/wsysmon-bin/wsysmon-bin.pacscript
0x262D: Fixed

Henryws:
> > Does this program have an icon?
> 
> This project does not have an icon, but I found a similar project with an icon that could be used. Is it worth it to do so?
> 
> Icon [https://github.com/KrispyCamel4u/SysMonTask/blob/master/icons/SysMonTask.png](https://github.com/KrispyCamel4u/SysMonTask/blob/master/icons/SysMonTask.png?rgh-link-date=2022-08-28T13%3A52%3A51Z)

If you want, sure. However I'd recommend not using other projects icons.

Henryws (approved):

PR #1538, opened 2022-08-31
add: `i3blocks-git`

laniusone:


Henryws (changes requested):
image

laniusone:
That's superweird, because I don't have `pkconf` installed, but it built properly on my system. Although it might be because I use zsh as user shell, so the check was skipped.

packages/i3blocks-git/i3blocks-git.pacscript
```diff
@@ -0,0 +1,26 @@
+name="i3blocks-git"
+pkgname="i3blocks"
+pkgdir="/usr/src/pacstall/${name}"
```
Henryws, line 3:
```suggestion
pkgdir="${STOWDIR}/${name}"
```

Henryws (changes requested):

packages/i3blocks-git/i3blocks-git.pacscript
```diff
@@ -0,0 +1,26 @@
...
+    git ls-remote "${url}" master | cut -f1 | cut -c1-8
+}
+version="$(pkgver)"
+
+prepare() {
+  ./autogen.sh && ./configure
+}
+
+build() {
+  make
```
Henryws, line 20:
```suggestion
  make -j"$(nproc)"
```

Henryws (changes requested):

packages/i3blocks-git/i3blocks-git.pacscript
```diff
@@ -0,0 +1,26 @@
...
+  ./autogen.sh && ./configure
+}
+
+build() {
+  make -j"$(nproc)"
+}
+
+install() {
+  # Install i3blocks
+  sudo make install DESTDIR=$pkgdir
```
Henryws, line 25:
```suggestion
  sudo make install DESTDIR="$pkgdir"
```

Henryws (changes requested):

packages/i3blocks-git/i3blocks-git.pacscript
```diff
@@ -0,0 +1,26 @@
+name="i3blocks-git"
+pkgname="i3blocks"
+pkgdir="${STOWDIR}/${name}"
+description="A feed generator for text based status bars"
+url="https://github.com/vivien/${pkgname}.git"
+depends="autoconf gcc make pkgconf"
```
Henryws, line 6: These aren't dependencies that this program needs at runtime, these are only needed to compile right?

Henryws (changes requested):

packages/i3blocks-git/i3blocks-git.pacscript
laniusone: You're right, changing it to `build_depends`

Henryws (approved):

PR #1548, opened 2022-09-07
Remove mirenradia as maintainer

mirenradia:
Unfortunately, I have switched to a non-Debian based distro so no longer use pacstall. Since I can no longer test packages, locally, I would like to remove myself as a maintainer.


Henryws (approved):

Henryws:
I'll take over `bat-deb`.

PR #1604, opened 2022-10-01
upd(pipewire): `0.3.54` -> `0.3.59`

Henryws:


Henryws:
Ok the issue with `develop` is a bug relating to `unzip`. The issue is fixed in https://github.com/pacstall/pacstall/pull/643

Henryws:
Debian stable is the only failure, everything else works.

PR #1606, opened 2022-10-02
upd(batsignal): `1.5.1` -> `1.6.2`

harsh18262:


harsh18262:
Hi, I just want to confirm is there a issue with some versions of pacstall that some tests have not passed or is it my changes in the pacscript, because the issue in the tests is that libnotify-dev package is not getting installed by pacstall

Henryws:
https://github.com/Arkq/cmusfm/issues/28

Try installing these packages

harsh18262:
Hi, I tried installing the packages but i think pacstall is not installing the dependencies in the some tests environments. I tried to install the package in a docker container with debian bullseye image the install worked. it just showed the error of libcurl4-openssl-dev not being found which is not even required so we can remove it

Henryws:
Alright

PR #1671, opened 2022-10-20
add: `ttf-fira-sans`

heyzsh:


Henryws:
Split adding packages into separate PRs.

PR #1677, opened 2022-10-21
fix(android-studio): delete symlink in removescript

Zahrun:


Zahrun:
On a side note, should this package be called "android-studio-bin"? Since it is not compiled but only binaries installed.

Henryws:
No. `-bin` packages contain a single binary, and sometimes a desktop file. This is more complex, so it is not a `-bin` package.

Henryws (approved):

Zahrun:
> No. `-bin` packages contain a single binary, and sometimes a desktop file. This is more complex, so it is not a `-bin` package.

I see. Thank you for the explanation

PR #1682, opened 2022-10-22
fix(obs-backgroundremoval-git): fix install for pacstall 3

Zahrun:


packages/obs-backgroundremoval-git/obs-backgroundremoval-git.pacscript
```diff
@@ -37,6 +37,7 @@ build() {
 }
 
 install() {
+  cd build || exit
```
Henryws, line 40:
```suggestion
  cd build
```

Henryws (changes requested):

packages/obs-backgroundremoval-git/obs-backgroundremoval-git.pacscript
Zahrun: image

Henryws: I am aware. Functions run in an environment where any non-zero exit code fails out completely. So saying `cd build || exit` is redundant.

Zahrun: As you wish. I now learned the `git commit --no-verify` option to by-pass the pre-commit hooks.

Henryws:
CI fails because it can’t `cd build`

Zahrun:
> CI fails because it can’t `cd build`

I think pacstall 3.0.1 changed the behavior. Can close this.

Henryws:
> 

How? The only change is in unguarded variables.

Zahrun:
> 

In 3.0.0, when I cd to build directory in build(), install() would start on parent directory. In 3.0.1, install() starts in same directory left by build()

Henryws:
You’re saying it would reset the dir on every function?

Zahrun:
> You’re saying it would reset the dir on every function?

That’s the behavior I observed in 3.0.0

PR #1685, opened 2022-10-22
fix(tenacity-git): fix install for Ubuntu 22.10

Zahrun:


packages/tenacity-git/tenacity-git.pacscript
```diff
@@ -19,14 +19,15 @@ prepare() {
 
 build() {
     mkdir build
-    cd build
+    cd build || exit
```
Henryws, line 22:
```suggestion
    cd build
```

packages/tenacity-git/tenacity-git.pacscript
```diff
@@ -19,14 +19,15 @@ prepare() {
...
+        -DESTDIR="${STOWDIR}/${name}" \
         -DCMAKE_BUILD_TYPE=Release \
         -Duse_ffmpeg=loaded ..
-    sudo make -j$(nproc)
+    sudo make -j"$(nproc)"
 }
 
 install() {
-    sudo make install DESTDIR="$STOWDIR/$name"
+    cd build || exit
```
Henryws, line 31:
```suggestion
    cd build
```

Henryws (changes requested):

Henryws:
CI fails.

Zahrun:
Hum I am confused now. I compiled this yesterday fine and today can’t do it anymore.
@echometerain, any idea?

Henryws:
Tenacity appears to be archived (https://github.com/tenacityteam/tenacity)

packages/tenacity-git/tenacity-git.pacscript
```diff
@@ -21,12 +21,12 @@ build() {
     mkdir build
     cd build
     cmake -G "Unix Makefiles" \
-        -DESTDIR="$STOWDIR/$name" \
+        -DDESTDIR="${pkgdir}" \
         -DCMAKE_BUILD_TYPE=Release \
         -Duse_ffmpeg=loaded ..
```
echometerain, line 26:
```suggestion
        -DCMAKE_BUILD_TYPE=Release ..
```
try this maybe?

Zahrun: I have set -DFFMPEG=OFF and it compiles now. Is the ffmpeg loading important?
I think the successor will be Saucedacity, but in current state I prefer tenacity

Henryws: I think ffmpeg loading is important, as it can handle a lot of different audio codecs.

Zahrun: I figured out the aur package uses specific version of ffmpeg 4.4 and that ubuntu 22.10 updated from 4.4 to 5.1.
So I think there is no luck for kinetic users like me.
But compiles fine in jammy.
So I just say in case kinetic, compile without ffmpeg. Otherwise, compile with ffmpeg
Unless and until someone creates a ffmpeg 4.4 pacscript for kinetic, I think that will do.

Zahrun:
Now installs fine on my side on jammy (with ffmpeg) and kinetic (no ffmpeg). The CI builds fine too, but for some reason fails to install.

Henryws:
It builds fine but if it can’t install, it means nothing.

Zahrun:
> It builds fine but if it can’t install, it means nothing.

Now it installs

PR #1717, opened 2022-10-31
upd(mutt-wizard): `3.2.1` -> `3.3.1`

JonBoyleCoding:
Also added repology link


JonBoyleCoding:
Looking at the integration tests - there is an issue with installing msmtp in that it asks whether to install it with apparmor support. I am unsure of what to do here as this could be user-dependent on whether they wish to or not (although probably should...).

Any thoughts?

Henryws:
Send a screenshot of the working build then.

Henryws (approved):
Works locally

PR #1733, opened 2022-11-05
add: `os-installer-git`

cat-master21:


cat-master21:
works on ubuntu Kinetic and Debian Sid only https://github.com/cat-master21/pacstall-programs/actions/runs/3401189735/jobs/5656130607 due to libgnome-desktop having different library names.

Henryws:
Can you make a case statement for each distros depends?

cat-master21:
I'd rather not as the amount of distros that it would support keeps changing (like adding lunar and bookworm to stable soon). If the package isn't avaible then it would just fail anyway @Henryws.

Henryws:
Ok then at least add a warning that tells the user that this package is intended for a very specific use-case, and should not be installed unless they know why they need to.

packages/os-installer-git/os-installer-git.pacscript
```diff
@@ -0,0 +1,37 @@
...
+version="$(pkgver)"
+
+prepare() {
+  git submodule update --init --recursive
+  fancy_message warn "This can only be built using the latest version of most OSes."
+}
+
+build() {
+  # Build nessesary files
+  meson --prefix="${STOWDIR}/${name}/usr" --buildtype=release . build
```
Henryws, line 25:
```suggestion
  meson --prefix="${pkgdir}/usr" --buildtype=release . build
```

packages/os-installer-git/os-installer-git.pacscript
```diff
@@ -0,0 +1,37 @@
...
+
+install() {
+  # Install license
+  sudo install -Dm644 "LICENSE" doc/* -t "${pkgdir}/usr/share/licenses/${pkgname}"
+
+  # Install doc
+  sudo install -Dm644 "README.md" -t "${pkgdir}/usr/share/doc/${pkgname}"
+
+  # Install
+  sudo meson install -C build --destdir "${STOWDIR}/${name}/usr"
```
Henryws, line 36:
```suggestion
  sudo meson install -C build --destdir "${pkgdir}/usr"
```

Henryws (changes requested):

Henryws:
Hmm, could this actually be put into a separate repository, or could it stay here..

cat-master21:
Well I'd rather it be here so that it can be accessible to other users more easily and if your refering to Rhino Linux's repo I heard that we should put the pacscripts here as you are still maintianing both and it's just less work generally.

Henryws (approved):

PR #1782, opened 2022-11-24
add: `protonup-qt-git`

DavidoTek:
Adds a ProtonUp-Qt `pascript`. It will fetch and download the latest tag release from GitHub.
Issue for ProtonUp-Qt: https://github.com/DavidoTek/ProtonUp-Qt/issues/83

## Progress 

- [x] Edit packagelist
- [x] Add initial pacscript
- [X] Contact devs (me @DavidoTek)
- [ ] Add maintainer to pacscript


packages/protonup-qt-git/protonup-qt-git.pacscript
```diff
@@ -0,0 +1,39 @@
...
+#!/bin/bash
+PYTHONPATH=$PYTHONPATH:${STOWDIR}/${name}/usr/local/lib/`ls ${STOWDIR}/${name}/usr/local/lib/ | grep python3`/dist-packages/ python3 -m pupgui2
+EOL
+
+    sudo ln -s "${STOWDIR}/${name}/usr/bin/net.davidotek.pupgui2.sh" "/usr/bin/protonup-qt"
+}
+
+build() {
+}
+
```
Henryws, line 30:
```suggestion
```

Henryws (changes requested):

packages/protonup-qt-git/protonup-qt-git.pacscript
```diff
@@ -0,0 +1,36 @@
+name="protonup-qt-git"
+description="Install and manage Proton-GE and Luxtorpeda for Steam and Wine-GE for Lutris"
+url="https://github.com/DavidoTek/ProtonUp-Qt.git"
+
+depends="python3"
+build_depends="python3 python3-pip"
+
```
Henryws, line 7:
```suggestion
```

packages/protonup-qt-git/protonup-qt-git.pacscript
```diff
@@ -0,0 +1,36 @@
+name="protonup-qt-git"
+description="Install and manage Proton-GE and Luxtorpeda for Steam and Wine-GE for Lutris"
+url="https://github.com/DavidoTek/ProtonUp-Qt.git"
+
```
Henryws, line 4:
```suggestion
```

packages/protonup-qt-git/protonup-qt-git.pacscript
```diff
@@ -0,0 +1,36 @@
...
+    sed 's/==.*//g' requirements.txt > requirements_nover.txt
+    pip3 install --verbose --no-index --ignore-installed --prefix="${STOWDIR}/${name}/usr" -r requirements_nover.txt
+
+cat >"${STOWDIR}/${name}/usr/bin/net.davidotek.pupgui2.sh" <<EOL
+#!/bin/bash
+PYTHONPATH=$PYTHONPATH:${STOWDIR}/${name}/usr/local/lib/`ls ${STOWDIR}/${name}/usr/local/lib/ | grep python3`/dist-packages/ python3 -m pupgui2
+EOL
+
+    sudo ln -s "${STOWDIR}/${name}/usr/bin/net.davidotek.pupgui2.sh" "/usr/bin/protonup-qt"
```
Henryws, line 25: This won't work. You probably want `sudo ln -s /usr/bin/net.davidotek.pupgui2.sh /usr/bin/protonup-qt` in `postinst`

Henryws (changes requested):

Henryws:
CI fails.

DavidoTek:
I see. Is it possible to install Python dependencies using `pip`? PySide6 isn't available in the Ubuntu repos.

```
Using pip 22.0.2 from /usr/lib/python3/dist-packages/pip (python 3.10)
ERROR: Could not find a version that satisfies the requirement PySide6-Essentials>=6.3.0 (from versions: none)
ERROR: No matching distribution found for PySide6-Essentials>=6.3.0
[!] ERROR: Could not prepare protonup-qt-git properly
```

Henryws:
Sure.

Henryws:
@DavidoTek any updates on this?

DavidoTek:
I did some investigation. Seems like `PySide6-Essentials` isn't available for the used Python version. `PySide6` should work fine.

DavidoTek:
Hmm, not sure what's causing the issue.
Do you have any idea what the problem is? Maybe I am missing something obvious?

sonic2kk:
> Using pip 22.0.2 from /usr/lib/python3/dist-packages/pip (python 3.10)
> ERROR: Could not find a version that satisfies the requirement PySide6 (from versions: none)
> ERROR: No matching distribution found for PySide6

These errors usually come from a compatible package version not being found for the current Python version (in this case, Python 3.10). But there should be a `pyside6` package available (afaik Pip packages are case-insensitive so casing shouldn't matter). The supported versions noted on the [PyPI](https://pypi.org/project/PySide6/) page should work with Python 3.10 I believe too: `Python <3.12, >=3.7`.

I'm not sure if this could be causing issues, but the PyPI page also notes:

> Please note: this wheel is an alias to other two wheels [PySide6_Essentials](https://pypi.org/project/PySide6_Essentials) and [PySide6_Addons](https://pypi.org/project/PySide6_Addons), which contains a predefined list of Qt Modules.

DavidoTek:
> These errors usually come from a compatible package version not being found for the current Python version 

That's what I though first too, there was a similar issue with the AppImage (that's why it's still using PySide 6.2).
But I don't see a reason why this should occur here, the Python version is new enough.

> Please note: this wheel is an alias to other two wheels [PySide6_Essentials](https://pypi.org/project/PySide6_Essentials) and [PySide6_Addons](https://pypi.org/project/PySide6_Addons)

Problem occurs with both `PySide6` and `PySide6-Essentials` (versions gets removed from the requirements.txt before installing).

Henryws:
@DavidoTek Any updates?

DavidoTek:
> @DavidoTek Any updates?

I've no clue why it is not working. For some reason, installing PySide6 using pip doesn't work. I'm also not sure if it is a good idea to install/remove global Python dependencies as it is not that simple to detect them, so maybe a different approach would be better:

**Alternatively**, I could create a different new *pacscript* that downloads and installs the [AppImage](https://github.com/DavidoTek/ProtonUp-Qt/releases) instead. Like the AUR package [protonup-qt-bin](https://aur.archlinux.org/packages/protonup-qt-bin).
That would also eliminate the problem with the global dependencies as all dependencies are included inside the AppImage.
The AppImage is ~58MB in size. Would it be okay to do it this way @Henryws ?

wizard-28:
> > @DavidoTek Any updates?
> 
> I've no clue why it is not working. For some reason, installing PySide6 using pip doesn't work. I'm also not sure if it is a good idea to install/remove global Python dependencies as it is not that simple to detect them, so maybe a different approach would be better:
> 
> **Alternatively**, I could create a different new _pacscript_ that downloads and installs the [AppImage](https://github.com/DavidoTek/ProtonUp-Qt/releases) instead. Like the AUR package [protonup-qt-bin](https://aur.archlinux.org/packages/protonup-qt-bin). That would also eliminate the problem with the global dependencies as all dependencies are included inside the AppImage. The AppImage is ~58MB in size. Would it be okay to do it this way @Henryws ?

Yeah making an AppImage pacscript is fine, name it `protonup-qt-app` as per our convention. Leave this PR open for now and make a new PR.

Henryws:
`protonup-qt-app` was added so closing.

PR #1800, opened 2022-12-07
upd(helix): `22.08.1` -> `22.12`

tranzystorek-io:


Henryws (changes requested):
add feature named-profiles. That's what cargo says on the debian runs

Henryws (approved):

PR #1838, opened 2022-12-14
add: `pulsar-deb`

Zahrun:


Zahrun:
For information, on my system, I need to run it as `pulsar --in-process-gpu`, otherwise it crashes, but that’s a known bug

wizard-28 (approved):

PR #1841, opened 2022-12-16
ci(tests): add step to uninstall packages

wizard-28:


.github/workflows/tests.yml
```diff
@@ -88,6 +88,25 @@ jobs:
...
+          export DEBIAN_FRONTEND=noninteractive
+          export GITHUB_ACTIONS=true
+          for changed_file in ${{ steps.files.outputs.added_modified }}; do
+            if [[ ${changed_file} == *".pacscript" ]]; then
+              pacscript_file=`basename "${changed_file}"`
+              package_name="${pacscript_file/.pacscript/ }"
+              package_dir=`dirname "${changed_file}"`
+              echo "Running pacstall -R for ${package_name}..."
+              cd ${package_dir}
+              pacstall --disable-prompts -R ${pacscript_file}
```
Henryws, line 104: Is this the pacscript variable without `.pacscript`?

wizard-28: No, I think `package_name` would be the correct variable. You can suggest the correct code.

Henryws, line 104:
```suggestion
              pacstall --disable-prompts -R ${package_name}
```

.github/workflows/tests.yml
```diff
@@ -166,3 +185,22 @@ jobs:
...
```
Henryws, line 201:
```suggestion
              pacstall --disable-prompts -R ${package_name}
```

Henryws (changes requested):

Henryws (approved):

PR #2016, opened 2022-12-28
rm: `windscribe-alpha-gui-deb`

WRM-42:


wizard-28:
Why is this being removed?

PR #2101, opened 2023-01-04
fix(obs-backgroundremoval): fix for pacstall 3.1.1

Zahrun:


MisterDerpie (approved):
As per https://github.com/royshil/obs-backgroundremoval/issues/34#issuecomment-1370771959 fixes the issue that `LICENSE` is not found.

Henryws (approved):

Issue #2105, opened 2023-01-04
Run debian:testing and debian:unstable when debian:stable is incompatible


Zahrun:
Run debian:testing and debian:unstable CI integrations when debian:stable is incompatible
Currently it does not, eg. https://github.com/pacstall/pacstall-programs/pull/2104

wizard-28:
Oh that's just the CI cancelling other runs when one of the Debian flavors fail. I don't think we can fix it.

PR #2108, opened 2023-01-04
ci: do not fail out if 1 package fails for debian too

Zahrun:
cf. https://github.com/pacstall/pacstall-programs/pull/1878
fixes https://github.com/pacstall/pacstall-programs/issues/2105


wizard-28 (approved):
Hmm. I thought @Henryws had already done this. Anyways LGTM.

PR #2151, opened 2023-01-07
ref(font-downloader): Remove `libgtk-3-dev` runtime dep

soumyatheman:


packages/font-downloader/font-downloader.pacscript
```diff
@@ -3,7 +3,7 @@ gives="font-downloader"
 version="10.0.0"
 build_depends="meson libhandy-1-dev gettext"
 # HACK: https://github.com/pacstall/pacstall/issues/727
```
Henryws, line 5:
```suggestion
```
You are not using any build_depends that don't exist in depends.

Henryws (changes requested):

packages/font-downloader/font-downloader.pacscript
soumyatheman: Actually that got copied from my template ig. Also, that's for a future confirmation. As libhandy-1-dev is a runtime dev too.

Henryws (approved):

PR #2212, opened 2023-01-16
add: `howdy-deb`

soumyatheman:


soumyatheman:
Is it not possible to pack this one as a pacscript?

Henryws:
It is, but the issue is it's asking for user input which we can't provide in a `-B` environment

PR #2250, opened 2023-01-19
add: `system76-power-deb`

soumyatheman:


packages/system76-power-deb/system76-power-deb.pacscript
```diff
@@ -0,0 +1,10 @@
+name="system76-power-deb"
+gives="system76-power"
+breaks="${gives}"
+version="1.1.23"
```
wizard-28, line 4: Outdated version, `1.1.24` is the latest.

wizard-28 (changes requested):

packages/system76-power-deb/system76-power-deb.pacscript
soumyatheman: Not available in the ppa

wizard-28: Get it from here I guess: https://apt-origin.pop-os.org/release/pool/jammy/system76-power/cee2d561326176396832fa4c1d3edf61556d2dba/

soumyatheman:
It also doesn't works in debian, I guess that may be the reason for not being in the repo.

wizard-28 (changes requested):
Make it `incompatible` with debian bullseye

PR #2479, opened 2023-02-18
add: `muse-hub`

huetremy:
First, thank you for this great tool !
I'd like to add MuseScore (a great free music edition tool) to pacstall as it is only present as 2.x and 3.4.x in apt, where last version is 4.X with great improvements.
`muse-hub` would be an optional dependency of MuseScore.

Please note that I added `icu-devtools` as a dependency because `muse-hub` seems to need `libicu`.
However, this library is only provided along with a version number in apt, and that depends on the ubuntu/debian version.
I noticed that there is a `libicu71-deb` package in pacstall. Would it be better to use it ?

Moreover, I'd like some help with the `removescript`: despite the `rm -rf` I did in it, the directory `/srv/muse-hub` still exists after uninstall (I tried using both `apt remove` and `pacstall -R`)

Thanks!

### DIFF
--- a/packagelist
+++ b/packagelist
@@ -208,6 +208,7 @@ moka-icon-theme
 ms-teams-deb
 mullvad-vpn-beta-deb
 mullvad-vpn-deb
+muse-hub-deb
 mutt-wizard
 nala-deb
 nano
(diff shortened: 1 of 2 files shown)


Henryws:
> I noticed that there is a libicu71-deb package in pacstall. Would it be better to use it ?

Depends if it's newer than what the apt equivalent is

huetremy:
> Depends if it's newer than what the apt equivalent is

Same version on my ubuntu 22.10 (and 23.04 for now), newer for other versions and debian

Henryws (approved):

PR #2488, opened 2023-02-20
add: `codium-deb`

oklopfer:


Henryws:
Already exists in https://github.com/pacstall/pacstall-programs/blob/master/packages/vscodium-deb/vscodium-deb.pacscript

PR #2535, opened 2023-02-24
add: `boosteroid-deb`

DismissedGuy:
Is the pacscript fine like this? Boosteroid only provides a generic download url, so it will cause a hash mismatch if they release an update...


Henryws:
Is there anywhere that provides a versioned URL?

DismissedGuy:
I spent some time hunting for it but couldn't find any. Even the version log the client downloads just seems to point to a generic `Bin.7z`. The [AUR package](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=boosteroid) also points to the unversioned one, so I don't think any exist.

Henryws:
Then I don't think that this package should become a pacscript if it can't provide a versioned URL.

DismissedGuy:
How about using `pkgver` to essentially make it autoupdate? The wiki is a bit contradicting on whether to allow pkgver for non-git packages though.

Henryws:
> The wiki is a bit contradicting on whether to allow pkgver for non-git packages though.

Don't use it for non-git packages.

DismissedGuy:
Understood. In that case it would indeed be impossible to realize this as a proper pacscript.

PR #2730, opened 2023-03-19
upd(pap): `0.13.4` -> `0.14.1`

talwat:


talwat:
Fixes #2729

talwat:
Debian fails because its version of golang is too old, so it was disabled on debian, and because of this it returns a non-0 exit code so the CI reports it as failed.

Henryws (approved):

PR #2732, opened 2023-03-19
add: `docker-bin`

oklopfer:


packages/docker-bin/docker-bin.pacscript
```diff
@@ -0,0 +1,29 @@
+name="docker-bin"
+gives="docker"
+replace="${gives}"
+repology=("project: ${gives}")
+version="23.0.1"
+depends="procps git iptables"
+description="Docker Engine Static Binaries"
+
+if [ $(dpkg --print-architecture) = arm64 ]; then
```
Henryws, line 9:
```suggestion
if [[ $CARCH == "amd64" ]]; then
```

Henryws (changes requested):

packages/docker-bin/docker-bin.pacscript
oklopfer: Do you mean $CARCH == arm64 or $CARCH !== amd64

Henryws: If `$CARCH == 'amd64'`

oklopfer: why tho

Henryws (approved):

PR #2790, opened 2023-03-21
add: `cmake-deb`

Zahrun:
Did you see this coming?

### DIFF
--- a/packages/cmake-deb/cmake-deb.pacscript
+++ b/packages/cmake-deb/cmake-deb.pacscript
@@ -7,7 +7,7 @@ codename="$(lsb_release -cs)"
 case "${codename}" in
   "bullseye")
     url="http://ftp.debian.org/debian/pool/main/c/cmake/cmake_${version}-1~bpo11+1_amd64.deb"
-    hash="6c7c8d91c045ccf44c9cc7dc1f05d8431707df4651bf18b4a3ee34847bc205f3"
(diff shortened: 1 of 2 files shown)


packages/cmake-deb/cmake-deb.pacscript
```diff
@@ -0,0 +1,19 @@
+name="cmake-deb"
+gives="cmake"
+version="3.25.1"
+description="CMake data files (modules, templates and documentation)"
+pacdeps=("cmake-data-deb")
+codename="$(lsb_release -cs)"
+case ${codename} in
```
Henryws, line 7:
```suggestion
case "${codename}" in
```

Henryws (changes requested):

Henryws (approved):

Zahrun:
@Henryws 
```
CMake Error at CMakeLists.txt:1 (cmake_minimum_required):
  CMake 3.25.2 or higher is required.  You are running version 3.25.1
```
😂
I guess I will wait for the patch update then 😅

Henryws:
Ok, make a new PR, and add `incompatible` to it.

Zahrun:
No, I mean I wanted to use this updated `cmake` to update `obs-backgroundremoval-git`, but it requires an ever so slightly newer cmake 😅
It's ok, I guess debian and ubuntu will catch up in some time. then I'll update these pacscripts to 3.25.2

PR #2796, opened 2023-03-23
upd(zotero): `6.0.22` -> `6.0.23`

JonBoyleCoding:


JonBoyleCoding:
The failure appears to be in the debian build - it cannot seem to find pacstall, so I don't think this is an issue on my part or for this PR.

PR #3245, opened 2023-05-05
add: `kicad-diff-git`

MarcelRobitaille:
## Progress 

- [x] Edit packagelist
- [x] Add initial pacscript
- [ ] Contact devs
- [ ] Add maintainer to pacscript


packages/kicad-diff-git/kicad-diff-git.pacscript
```diff
@@ -0,0 +1,31 @@
+name="kicad-diff-git"
+repology=("project: kicad-diff-git")
+version="4500fef3"
```
Henryws, line 3:
```suggestion
```

Henryws (changes requested):

MarcelRobitaille:
@Henryws thanks for reviewing. Is this what you meant?

packages/kicad-diff-git/kicad-diff-git.pacscript
```diff
@@ -0,0 +1,31 @@
...
+build_depends=("git" "python3-pip")
+depends=("python3")
+gives="kicad-diff"
+description=" Scripts for performing image diffs between pcbnew layout revisions Resources"
+maintainer="Marcel Robitaille <mail@marcelrobitaille.me>"
+arch=("any")
+provides=("kicad-diff")
+conflicts=("kicad-diff")
+url="https://github.com/leoheck/KiCad-Diff.git"
+
```
Henryws, line 14:
```suggestion
```

Henryws (changes requested):

packages/kicad-diff-git/kicad-diff-git.pacscript
MarcelRobitaille: Isn't `url` required? It's used in `pkgver()` and to download the source code, right? If I try to remove it, I get:

```
(kicad-diff-git) Do you want to view/edit the pacscript [y/N]
[+] INFO: Sourcing pacscript
fatal: no path specified; see 'git help pull' for valid url syntax
fatal: no path specified; see 'git help pull' for valid url syntax
[!] ERROR: Package does not contain version
```

Henryws: I said to to remove the blank line, not URL.

MarcelRobitaille: Oh sorry the diff in GitHub makes it look like I should replace 11-14 with just an empty line

packages/kicad-diff-git/kicad-diff-git.pacscript
```diff
@@ -0,0 +1,30 @@
+name="kicad-diff-git"
+repology=("project: kicad-diff-git")
+epoch="1"
```
Henryws, line 3: Why is epoch needed?

Henryws (changes requested):

packages/kicad-diff-git/kicad-diff-git.pacscript
```diff
@@ -0,0 +1,29 @@
+name="kicad-diff-git"
+repology=("project: kicad-diff-git")
+homepage="https://github.com/leoheck/KiCad-Diff"
+build_depends=("git" "python3-pip")
+depends=("python3")
+gives="kicad-diff"
+description=" Scripts for performing image diffs between pcbnew layout revisions Resources"
```
Henryws, line 7:
```suggestion
description="Scripts for performing image diffs between pcbnew layout revisions Resources"
```

packages/kicad-diff-git/kicad-diff-git.pacscript
```diff
@@ -0,0 +1,29 @@
+name="kicad-diff-git"
+repology=("project: kicad-diff-git")
+homepage="https://github.com/leoheck/KiCad-Diff"
+build_depends=("git" "python3-pip")
```
Henryws, line 4:
```suggestion
build_depends=("python3-pip")
```

Henryws (changes requested):

Henryws:
What should we do about this?

MarcelRobitaille:
@Henryws Sorry what should we do about what

Henryws:
The CI failing.

MarcelRobitaille:
Seems to be an error resulting from `pip install wxpython`:
```
ModuleNotFoundError: No module named 'attrdict'
```

The solution I've found online is to use Python 3.10. For things like this, I usually use pyenv.

Henryws:
So why can't we use 3.10?

MarcelRobitaille:
It probably has to do with the packaged version in Ubuntu. Maybe the pacstall script worked on my computer because I had pyenv, which injects its own version of python into $PATH. I'm traveling right now. Maybe when I'm back I could figure out a script with pyenv?

Elsie19:
Stale.

PR #3295, opened 2023-05-07
refactor(*-git): use timestamped versioning

oklopfer:
This proposed change would solve the issue of Pacstall `git` packages needing to be downgraded, due to them being versioned only by `commit_hash`, which is not necessarily incremental and conflicts with how dpkg/apt are meant to function, meaning Pacstall has to force downgrade `git` packages. 

To fix this, the new `pkgver()` calculation will now produce `commit_UNIX_timestamp.commit_hash`. As an example:

`emacs-git` is currently versioned `b625ccff`, giving the commit hash

the updated versioning of `emacs-git` would be `1683575994.b625ccf`, giving the UNIX timecode for when the commit was created, with a dot version of the commit hash

This would allow for us to still identify the packages with the hash, but conform to Debian systems and update packages incrementally, properly.


oklopfer:
@Zahrun @echometerain @DiegoMagdaleno @kwon37xi @ObsidianPresidium @laniusone @Ruturajn @edward-riley You have (a) Pacstall package(s) that will be affected by this change. Could you reply to this comment, indicating if you give consent to this change? Thank you.

Zahrun:
This looks like a lot of repetition of code. Could we instead have this done by pacstall for all -git packages?

DiegoMagdaleno:
@oklopfer I consent.

oklopfer:
> This looks like a lot of repetition of code. Could we instead have this done by pacstall for all -git packages?

We discussed a backend way of doing this, but realized issues would arise for -git packages that don't use this style versioning. This is the only way we can do it without making a "breaking change"

Zahrun:
> > This looks like a lot of repetition of code. Could we instead have this done by pacstall for all -git packages?
> 
> We discussed a backend way of doing this, but realized issues would arise for -git packages that don't use this style versioning. This is the only way we can do it without making a "breaking change"

Ok, go ahead then

Ruturajn:
I consent to this change.

Thank You

echometerain:
I consent, thanks

Henryws:
> > ?
> 
> We discussed a backend way of doing this, but realized issues would arise for -git packages that don't use this style versioning. This is the only way we can do it without making a "breaking change"

This might be turned into a Pacstall function later that people can opt into after this is merged.

oklopfer:
Closed in favor of pacstall 4.0.0

PR #3306, opened 2023-05-08
upd(obs-backgroundremoval): `0.5.16` -> `0.5.19`

Zahrun:
Requires https://github.com/pacstall/pacstall-programs/pull/3304


Henryws:
This fails on all the CI's

Zahrun:
> This fails on all the CI's

Yes, it required https://github.com/pacstall/pacstall-programs/pull/3304
Now it fails on cmake-deb uninstall but the install is fine

PR #3431, opened 2023-05-14
add: `papirus-folders`

oklopfer:


packages/papirus-folders/papirus-folders.pacscript
```diff
@@ -0,0 +1,14 @@
+name="papirus-folders"
+gives="${name}"
+version="1.12.1"
+description="Papirus folder theme"
+repology=("project: papirus-folders")
+depends=("papirus-icon-theme")
+build_depends=("make")
```
Henryws, line 7:
```suggestion
```
That is already installed with pacstall.

packages/papirus-folders/papirus-folders.pacscript
```diff
@@ -0,0 +1,14 @@
+name="papirus-folders"
+gives="${name}"
```
Henryws, line 2:
```suggestion
```
If $name exists, $gives implicitly sets to $name unless otherwise specified.

Henryws (changes requested):

Henryws:
😁

Henryws (approved):

PR #3575, opened 2023-05-23
fix(linux-kernel): repology

oklopfer:


wizard-28 (approved):
`repo` filter may be overkill, but better be safe than sorry.

PR #3660, opened 2023-06-07
add: `foliate-deb`

Diegiwg:
## Progress 

- [x] Edit packagelist
- [x] Add initial pacscript
- [x] [Contact devs](https://github.com/johnfactotum/foliate/issues/1041)

### DIFF
--- a/packages/foliate-deb/foliate-deb.pacscript
+++ b/packages/foliate-deb/foliate-deb.pacscript
@@ -1,5 +1,4 @@
 name="foliate-deb"
-gives="foliate"
 version="2.6.4"
 arch=("any")
 url="https://github.com/johnfactotum/foliate/releases/download/${version}/com.github.johnfactotum.foliate_${version}_all.deb"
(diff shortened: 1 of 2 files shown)


Diegiwg:
`foliate` accepted, and asked to make a PR publishing the installation method (which I already did).

I used the url that I believe will be the `final` one when this package is published in the expo PR.

Waiting for the package to be published to check for any errors, and correct it in the foliate documentation.

packages/foliate-deb/foliate-deb.pacscript
```diff
@@ -0,0 +1,8 @@
+name="foliate-deb"
```
Henryws, line 1: Add `gives`.

Henryws (changes requested):

packages/foliate-deb/foliate-deb.pacscript
Henryws: Why would `gives="com.github.johnfactotum.foliate"` not work?

Diegiwg: I had tested it with gives="com.github.johnfactotum.Foliate", and it didn't work.

I didn't realize that the `F` in `.Foliate` would break gives.

Henryws (approved):

PR #3743, opened 2023-06-10
add: `whatsie-git`

Diegiwg:
## Progress 

- [x] Edit packagelist
- [x] Add initial pacscript
- [x] [Contact devs](https://github.com/keshavbhatt/whatsie/issues/123)


Diegiwg:
I'm not understanding what is breaking. Here on my machine, you can do the entire installation process. 

;;

packages/whatsie-git/whatsie-git.pacscript
```diff
@@ -0,0 +1,19 @@
+name="whatsie-git"
+gives="whatsie"
+version="4.14.1"
```
Henryws, line 3: This should be using `pkgver()`. Consult the wiki and [other git packages](https://github.com/search?q=repo%3Apacstall%2Fpacstall-programs%20pkgver&type=code) on how to do this.

Henryws (changes requested):

Henryws:
> I'm not understanding what is breaking. Here on my machine, you can do the entire installation process.
> 
> ;;

It was issues with our CI. It should be fixed now.

packages/whatsie-git/whatsie-git.pacscript
Diegiwg: I don't know much about shell, I got the list with the last output being the version I wanted with the following command: `git ls-remote "${url}" | grep -i "tags/v" | cut -d '/'  -f3 | cut -d 'v' -f2 | sort --version-sort`

Diegiwg: But then, it didn't work, returning the error:
```shell
[+] INFO: Packaging whatsie-git as whatsie
        [>] Packing control.tar
        [>] Packing data.tar
        [>] Compressing
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
Note, selecting 'whatsie:none' instead of '/usr/src/pacstall/whatsie-git.deb'
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 whatsie:none : Depends: qtbase5-dev:none but it is not installable
                Depends: qtdeclarative5-dev:none but it is not installable
                Depends: qtlocation5-dev:none but it is not installable
                Depends: libqt5webchannel5-dev:none but it is not installable
                Depends: qtwebengine5-dev:none but it is not installable
                Depends: qtwebengine5-private-dev:none but it is not installable
        [!] ERROR: Failed to install whatsie-git deb
[+] INFO: Cleaning up
```

Diegiwg: I thought that if I gave pacscript a list, it would get the first result automatically.

But that didn't happen, so he didn't know which version to use. Then, I looked for how to get the first line of a shell output and that's it, solved.

`git ls-remote "${url}" | grep -i "tags/v" | cut -d '/' -f3 | cut -d 'v' -f2 | sort -r --version-sort | head -n 1`

packages/whatsie-git/whatsie-git.pacscript
```diff
@@ -0,0 +1,23 @@
...
+arch=("amd64")
+url="https://github.com/keshavbhatt/whatsie.git"
+homepage="https://github.com/keshavbhatt/whatsie"
+description="Feature rich WhatsApp Client for Desktop Linux"
+maintainer="Diegiwg <diegiwg@gmail.com>"
+depends=("qtbase5-dev" "qtdeclarative5-dev" "qtlocation5-dev" "libqt5webchannel5-dev" "qtwebengine5-dev" "qtwebengine5-dev-tools" "qtwebengine5-private-dev" "libqt5webengine-data")
+build_depends=("qtbase5-dev" "qtdeclarative5-dev" "qtlocation5-dev" "libqt5webchannel5-dev" "qtwebengine5-dev" "qtwebengine5-dev-tools" "qtwebengine5-private-dev" "libqt5webengine-data")
+
+pkgver() {
+  git ls-remote "${url}" | grep -i "tags/v" | cut -d '/' -f3 | cut -d 'v' -f2 | sort -r --version-sort | head -n 1
```
Henryws, line 12:
```suggestion
  git ls-remote "${url}" main | head -n1 | cut -f1 | cut -c1-8
```

Henryws (changes requested):

packages/whatsie-git/whatsie-git.pacscript
```diff
@@ -0,0 +1,23 @@
+name="whatsie-git"
+gives="whatsie"
+arch=("amd64")
```
Henryws, line 3: This is a git package, so it should work on any architecture correct?

Diegiwg: I believe so, however, I haven't tested it on an arm, so I can't say.

Henryws: It's in C++, so remove the `arch`.

packages/whatsie-git/whatsie-git.pacscript
```diff
@@ -0,0 +1,23 @@
+name="whatsie-git"
+gives="whatsie"
+arch=("any")
```
Henryws, line 3:
```suggestion
```
If it works on everything, while "any" would *work*, removing it means the same.

Henryws (changes requested):

Henryws:
Can you make this as equal to whatsie as possible?

Diegiwg:
> Can you make this as equal to whatsie as possible?

Yes, as soon as I get home (+6 hours), I'll update the PR.

Henryws (approved):

PR #3751, opened 2023-06-11
refactor: `google-chrome-beta-deb`

Diegiwg:
I tried to automate version checking, I don't know if it's allowed, or I practice it. If not, you can just ignore it and let me know, and I will manually update the script.


Henryws:
No auto updates in a pacscript besides `-git` packages. Use `pacup` and `repology` variable for "auto updates"

PR #3765, opened 2023-06-12
chan(maitreya8-deb): change pkgname for repology

Zahrun:
Is it correct that this change would move pacstall to maitreya package in repology?
We are the only repo which has it under maitreya8 https://repology.org/project/maitreya8/versions
Whereas all other repos have it under maitreya https://repology.org/project/maitreya/versions

### DIFF
--- a/packages/maitreya8-deb/maitreya8-deb.pacscript
+++ b/packages/maitreya8-deb/maitreya8-deb.pacscript
@@ -1,6 +1,6 @@
 name="maitreya8-deb"
 repology=("project: maitreya")
-pkgname="maitreya8"
+pkgname="maitreya"
 version="8.0.1b"
 url="https://github.com/martin-pe/${pkgname}/releases/download/v${version}/${pkgname}_${version}_amd64.deb"
 depends=("libwxsqlite3-3.0-0" "wx-common wx3.0-i18n")


Henryws:
You'd have to change the package name as well.

Zahrun:
What do you think? Should we do it? Or is it useless?

Henryws:
When in Rome... Let's change it.

packages/maitreya-deb/maitreya-deb.pacscript
```diff
@@ -1,6 +1,6 @@
-name="maitreya8-deb"
+name="maitreya-deb"
 repology=("project: maitreya")
-pkgname="maitreya8"
+pkgname="maitreya"
```
Henryws, line 3:
```suggestion
```

packages/maitreya-deb/maitreya-deb.pacscript
```diff
@@ -1,6 +1,6 @@
...
 version="8.0.1b"
 url="https://github.com/martin-pe/${pkgname}/releases/download/v${version}/${pkgname}_${version}_amd64.deb"
```
Henryws, line 5:
```suggestion
url="https://github.com/martin-pe/maitreya/releases/download/v${version}/maitreya_${version}_amd64.deb"
```

Henryws (changes requested):

packages/maitreya-deb/maitreya-deb.pacscript
Zahrun: Wow quite an oversight, sorry

packages/maitreya-deb/maitreya-deb.pacscript
```diff
@@ -1,8 +1,7 @@
-name="maitreya8-deb"
+name="maitreya-deb"
 repology=("project: maitreya")
-pkgname="maitreya8"
 version="8.0.1b"
-url="https://github.com/martin-pe/${pkgname}/releases/download/v${version}/${pkgname}_${version}_amd64.deb"
+url="https://github.com/martin-pe/maitreya8/releases/download/v${version}/maitreya8_${version}_amd64.deb"
 depends=("libwxsqlite3-3.0-0" "wx-common wx3.0-i18n")
 replace=("maitreya8")
```
Henryws, line 6:
```suggestion
replace=("maitreya")
```

Henryws (changes requested):

packages/maitreya-deb/maitreya-deb.pacscript
Zahrun: I kept both names actually because the deb from github actually installs as maitreya8

Henryws (approved):

PR #3783, opened 2023-06-12
upd(libicu-deb): `71.1` -> `72.1`

Zahrun:


packages/libicu-deb/libicu-deb.pacscript
```diff
@@ -0,0 +1,9 @@
+name="libicu-deb"
+gives="libicu"
```
Henryws, line 2:
```suggestion
gives="libicu72"
```

Henryws (changes requested):

packages/libicu-deb/libicu-deb.pacscript
```diff
@@ -0,0 +1,9 @@
+name="libicu-deb"
+version="72.1"
+gives="libicu${version:0:2}"
+url="http://archive.ubuntu.com/ubuntu/pool/main/i/icu/libicu72_${version}-3ubuntu2_amd64.deb"
```
Henryws, line 4: You should probably add the string splicing to this as well

Henryws (changes requested):

packages/libicu-deb/libicu-deb.pacscript
Zahrun: Yeah but pacup fails to parse it :/
So anyway I would have to change it manually

Henryws: Oh true..

PR #3785, opened 2023-06-12
fix(fsearch-deb): new urls and hashes

Zahrun:


Zahrun:
debian:stable CI fails but I can’t reproduce

Henryws:
Fix the title.

Zahrun:
> Fix the title.

Version is still 0.2.2 though

Henryws:
Then why did you use the `upd` prefix?

Zahrun:
I guess it’s a fix then

PR #3809, opened 2023-06-19
upd(telegram-bin): `4.8.1` -> `4.8.4`

Diegiwg:
I'm not sure if it's allowed to switch maintainers, but if it's not, I'll be happy to revert the change, but still apply the update.

P.S.: The package is two commits away from being updated by third parties (other than the registered maintainer).


Henryws:
@wizard-28 are you ok for changing maintainership?

Henryws (approved):

PR #3843, opened 2023-06-26
rm: `nyxt-deb`

cat-master21:
The deb version of nyxt was removed in favor of a self contained package with the [`3.0.0` version](https://github.com/atlas-engineer/nyxt/releases/tag/3.0.0) which is very thick.


Henryws (changes requested):
Remove from packagelist

Henryws (approved):

PR #4434, opened 2023-08-26
add: `ungoogled-chromium-app`

rictonov:


Elsie19:
Do not unpack the appimage. Look at how other `-app` packages do it.

Elsie19:
Stale.

PR #4519, opened 2023-09-05
fix(lutris-deb): add `depends`

Xdavius:
Missing dependency from github .deb (useless if using OBS repository, bug reported to lutris team too)

Permit the header banner to be visible on lutris


Elsie19:
Fix the title.

Elsie19 (approved):

PR #4578, opened 2023-09-22
fix(obs-backgroundremoval-git): fix build scripts

umireon:


Zahrun (changes requested):

Zahrun (approved):
lgtm @Elsie19

Elsie19:
All the checks failed. Merge in from master

umireon:
I don't know why the uninstalling test fails. Does this test need to be passed?

Zahrun:
> I don't know why the uninstalling test fails. Does this test need to be passed?

This is only a CI issue due to the cmake-deb pacdep. We should be good to merge.

Elsie19 (approved):

PR #4589, opened 2023-09-26
upd(maitreya-deb): `8.0.1b` -> `8.1`

Zahrun:


Elsie19 (changes requested):
fails to remove.

Zahrun:
@Elsie19 will the change of `gives` change the name in repology? cf. https://github.com/pacstall/pacstall-programs/pull/3765

Elsie19:
Not that I know of.

Zahrun:
Ok this deb works only in jammy but it should be good to merge now

Elsie19 (approved):

PR #4651, opened 2023-10-11
add: `oh-my-posh-git`

Zahrun:


Elsie19 (changes requested):
CI fails.

Zahrun:
> CI fails.

For some reason, pacstall fails to install nerd-fonts-hack as a pacdep. If I install the pacdep manually with `pacstall -I nerd-fonts-hack`, then it works.

Zahrun:
The CI removes nerd-fonts-hack, which automatically removes oh-my-posh-git, and then tries to remove oh-my-posh-git, but fails because it already automatically was removed

oklopfer:
@Zahrun other than the CI's uninstall error, is your package all good to go?

Zahrun:
> @Zahrun other than the CI's uninstall error, is your package all good to go?

Yes

PR #4658, opened 2023-10-12
fix(oh-my-posh-git): works on arm

Zahrun:


Zahrun:
The CI removes nerd-fonts-hack, which automatically removes oh-my-posh-git, and then tries to remove oh-my-posh-git, but fails because it already automatically was removed

Elsie19 (approved):

PR #4684, opened 2023-10-20
fix(pacstall-gui-git): Change `settings.sh` install location

cat-master21:
In my script, the location of `settings.sh` has changed so I have updated this to reflect that.

### DIFF
--- a/packages/pacstall-gui-git/pacstall-gui-git.pacscript
+++ b/packages/pacstall-gui-git/pacstall-gui-git.pacscript
@@ -12,18 +12,18 @@ pkgver() {
 }
 
 prepare() {
-  sudo mkdir -p "${pkgdir}/usr/bin" "${pkgdir}$homedir/.config/pacstall-gui" "${pkgdir}/usr/share/applications"
+  sudo mkdir -p "${pkgdir}/usr/bin" "$homedir/.config/pacstall-gui" "${pkgdir}/usr/share/applications"
 }
 
 package() {
   sudo install -Dm755 pacstall-gui "${pkgdir}/usr/bin/pacstall-gui"
-  sudo mv -f ./settings.sh "${pkgdir}$homedir/.config/pacstall-gui"
+  sudo mv -f ./settings.sh "$homedir/.config/pacstall-gui"
   sudo install -Dm755 LICENSE "${pkgdir}/usr/share/licenses/pacstall-gui/LICENSE"
   sudo mv -f ./pacstall-gui.desktop "${pkgdir}/usr/share/applications"
 }
 
 post_install() {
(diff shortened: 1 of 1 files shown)


packages/pacstall-gui-git/pacstall-gui-git.pacscript
```diff
@@ -12,17 +12,18 @@ pkgver() {
 }
 
 prepare() {
-  sudo mkdir -p "${pkgdir}/usr/bin" "${pkgdir}/usr/lib/pacstall-gui" "${pkgdir}/usr/share/applications"
+  sudo mkdir -p "${pkgdir}/usr/bin" "${pkgdir}$HOME/.config/pacstall-gui" "${pkgdir}/usr/share/applications"
```
Elsie19, line 15: Don't use `$HOME`, use `$homedir`.

Elsie19 (changes requested):

Elsie19 (changes requested):
Actually where should we put these files, because we shouldn't be packing a Deb with hardcoded paths to home directories.

cat-master21:
In that case @Elsie19 I can just move it to the user's home directory. It is okay my program automatically regenerates the file in case it is missing.

packages/pacstall-gui-git/pacstall-gui-git.pacscript
```diff
@@ -12,17 +12,18 @@ pkgver() {
 }
 
 prepare() {
-  sudo mkdir -p "${pkgdir}/usr/bin" "${pkgdir}/usr/lib/pacstall-gui" "${pkgdir}/usr/share/applications"
+  sudo mkdir -p "${pkgdir}/usr/bin" "$homedir/.config/pacstall-gui" "${pkgdir}/usr/share/applications"
```
Elsie19, line 15: Do not put `$homedir` anything in any functions except post ones.

Elsie19 (changes requested):

packages/pacstall-gui-git/pacstall-gui-git.pacscript
cat-master21: Is the `settings.sh` file still available in post install if it is not put anywhere else though?

Elsie19 (commented):
Does this PR do anything? It seems to just be removing the creation of `"${pkgdir}/usr/lib/pacstall-gui"`, and removing `settings.sh`entirely.

cat-master21:
Yes, as in the script, `settings.sh`'s location got moved to `~/.config/pacstall-gui` but is not really needed because the script generates it if it is missing but it is not needed at `/usr/lib/pacstall-gui` anymore @Elsie19.

Elsie19 (approved):

PR #4731, opened 2023-10-30
fix(yad): change yad to yad-git

Xdavius:
Too many regressions/differences in yad github comparing to official yad 0.40 provided by debian/ubuntu

Packages must be differentiated to avoid problems.

### DIFF
--- a/packages/yad/yad-git.pacscript
+++ b/packages/yad/yad-git.pacscript
@@ -1,4 +1,4 @@
-name="yad"
+name="yad-git"
 pkgver="13.0"
 url="https://github.com/v1cont/${name}/releases/download/v${pkgver}/${name}-${pkgver}.tar.xz"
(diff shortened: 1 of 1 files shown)


D-Brox:
`steamtinkerlaunch-git` too

Elsie19:
This isn't a git package, so it shouldn't be `yad-git`.

Xdavius:
Hi.

It's a source release. It needs to be built.
How could be the good naming.

I had the same problem with discord  (discord, discord-deb, discord-bin) but discord doesn't exists in repositories of debian (I don't remember for Ubuntu)

I think packages who can conflicts need to be named differently.

Some propositions :

-current
-release
-src
-git-release
-src-release 

Archlinux doesn't have this problem because pacman is able to manage conflicts and compiling from source produce a pacman pkg.

Help will be appreciated.

Thank you.

Xdavius:
Aborting this change keeping the coherence for all packages

PR #4771, opened 2023-11-12
upd(docker-bin): `24.0.5` -> `24.0.7`

oklopfer:


packages/docker-bin/docker-bin.pacscript
```diff
@@ -2,24 +2,57 @@ name="docker-bin"
...
+Description=Docker Application Container Engine
+Documentation=https://docs.docker.com
+After=network-online.target docker.socket firewalld.service containerd.service time-set.target
+Wants=network-online.target containerd.service
+Requires=docker.socket
+
+[Service]
+Type=notify
+ExecStart=/usr/bin/dockerd -H fd:// --containerd=/run/containerd/containerd.sock
+ExecReload=/bin/kill -s HUP $MAINPID
```
Elsie19, line 36: since you used double quotes in the echo, this will be evaluated.

Elsie19 (changes requested):

packages/docker-bin/docker-bin.pacscript
```diff
@@ -2,24 +2,121 @@ name="docker-bin"
...
+# Only systemd 226 and above support this version.
+TasksMax=infinity
+OOMScoreAdjust=-999
+
+[Install]
+WantedBy=multi-user.target" | sudo tee "${pkgdir}/lib/systemd/system/containerd.service" > /dev/null
+}
+
+post_install() {
+  if [[ -z $(cat /etc/group | grep docker) ]]; then
```
Elsie19, line 108: use grep return codes instead of checking output.

Elsie19 (changes requested):

Elsie19 (approved):

PR #4927, opened 2023-11-19
add: `pb-git`

vigress8:


packages/pb/pb.pacscript
```diff
@@ -0,0 +1,17 @@
...
+pkgver="0.0.1"
+pkgdesc="0x0 pastebin utility"
+maintainer="vigress8 <vig@disroot.org>"
+url="https://git.envs.net/envs/pb.git"
+pkgver() {
+  git ls-remote "${url}" master | cut -f1 | cut -c1-8
+}
+
+prepare() {
+  git clone "${url}"
```
Elsie19, line 11: This isn't needed.

Elsie19 (changes requested):

packages/pb/pb.pacscript
vigress8: @Elsie19 when I remove `prepare()` I get this:

```
/tmp/pb.pacscript: line 11: cd: pb: No such file or directory
make: *** No rule to make target 'install'.  Stop.
	[!] ERROR: Could not package pb properly
```

packages/pb/pb.pacscript
```diff
@@ -0,0 +1,17 @@
...
+}
+
+package() {
+  cd "${name}"
```
Elsie19, line 15:
```suggestion
```

packages/pb/pb.pacscript
Elsie19: remove it and apply the other suggestion I made.

Elsie19 (changes requested):

packages/pb/pb.pacscript
vigress8: @Elsie19 exact same `make` error:
`No rule to make target 'install'.  Stop.`

Elsie19: ok, I will look later.

packages/pb-git/pb-git.pacscript
```diff
@@ -0,0 +1,12 @@
+name="pb-git"
+pkgver="0.0.1"
```
Elsie19, line 2:
```suggestion
pkgver="2022.11.03"
```

Elsie19 (changes requested):

Elsie19 (approved):

PR #4989, opened 2023-12-03
upd(youtube-music-deb): `2.2.0` -> `3.0.2`

Anifyuli:


Elsie19 (changes requested):
Why do you keep having branches with multiple things modified?

PR #4991, opened 2023-12-03
add: `scrcpy`

ghost:


packages/scrcpy/scrcpy.pacscript
```diff
@@ -0,0 +1,32 @@
+name="scrcpy"
+gives="${name}"
+pkgver="2.3.1"
+pkgdesc="Display and control your Android device"
+repology=("project: ${gives}")
+arch=("amd64")
```
oklopfer, line 6: any reason this wouldn't work on arm64?

vigress8: @oklopfer I suppose not. I couldn't find an official list of supported architectures, but seems like it's packaged for several arm64 distros

PR #5000, opened 2023-12-05
add: `kakoune`

vigress8:


Elsie19 (changes requested):
Add `repology`.

vigress8:
@Elsie19 it's there, line 4

PR #5009, opened 2023-12-06
refactor: replace deprecated `pkgname` with `gives`, remove redundant `gives` and `breaks`

vigress8:


oklopfer:
We are actually planning on switching back to `pkgname` in 5.0.0 for closer compatibility/comparability to PKGBUILDs, so we probably shouldn't remove it like this

vigress8:
@oklopfer bruh

PR #5021, opened 2023-12-11
upd(youtube-music-deb): `3.0.2` -> `3.1.0`

Anifyuli:


.gitignore
```diff
@@ -2,3 +2,4 @@
 *.tar.gz
 *.zip
 *.png
+pacstall-programs/*
```
oklopfer, line 5: remove this from the merge. this would break using `git add -A`

oklopfer (changes requested):

.gitignore
Anifyuli: So, I need to remove this? Or with other method?

oklopfer: correct, this line shouldn't be here. just go to https://github.com/Anifyuli/pacstall-programs/edit/ship-youtube-music-deb/.gitignore?pr=%2Fpacstall%2Fpacstall-programs%2Fpull%2F5021 and remove the line

oklopfer (approved):

PR #5058, opened 2023-12-19
upd(librewolf-app): `119.0.1-1` -> `120.0.1-1`

vigress8:


vigress8:
```/var/lib/dpkg/info/librewolf-app.postrm: line 60: return: can only `return' from a function or sourced script```
@oklopfer any clue what this is about?

**Comment by oklopfer** [comment]
Looks to be due to the previous maintainer's removal script integrating the `ask` function. We can probably rewrite it internally in the package.

**Comment by vigress8** [comment]
I removed `ask` and re-implemented the question, but the error persists. Is this an issue with `post_remove` in general?

**Comment by oklopfer** [comment]
@Elsie19 any ideas?

**Comment by Elsie19** [comment]
> ```/var/lib/dpkg/info/librewolf-app.postrm: line 60: return: can only `return' from a function or sourced script```
> @oklopfer any clue what this is about?

it means somewhere outside of a function it's running `return`, which bash doesn't like.

**Comment by vigress8** [comment]
@Elsie19 so how can this be fixed?

File: packages/librewolf-app/librewolf-app.pacscript [hunk 1]
```diff
@@ -1,37 +1,45 @@
...
+  amd64)
+    gnuarch="x86_64"
+    hash="21e73e405835bd6cfdaecc9930423f68e544eccb52deb7c992e90f5085ca487e"
+    ;;
+  arm64)
+    gnuarch="aarch64"
+    hash="af10277447c2c469d3b6c6a4ea99460de90346edfd8515066a317a954659e0f4"
+    ;;
+  *) return 1 ;;
+esac
```
oklopfer, line 17: revert this back to the `if` instead of the `case`. `arch=('amd64' 'arm64')` already ensures that there will be no `*)` case, and this is the only place in the script where we are using a return code. I have a feeling this is what is causing the issue.

oklopfer (changes requested):

packages/librewolf-app/librewolf-app.pacscript
vigress8: Giving up on `post_remove` for now. This issue needs to be looked into though

oklopfer:
https://github.com/pacstall/pacstall-programs/pull/5066
your changes were the problem. please stop unnecessarily changing variables to conform to your specific formatting standards when updating a pacscript. If only the has and version need updating, only the hash and version need updating.

vigress8:
I still can't figure out which exact change broke it

PR #5061, opened 2023-12-19
add: `megabasterd-bin`

vigress8:


Elsie19 (changes requested):
Add to `packagelist`.

packages/megabasterd-bin/megabasterd-bin.pacscript
```diff
@@ -0,0 +1,39 @@
+name="megabasterd-bin"
+gives="megabasterd"
+_name="MegaBasterd"
```
Elsie19, line 3: Remove that and just use the actual string instead.

Elsie19 (changes requested):

packages/megabasterd-bin/megabasterd-bin.pacscript
```diff
@@ -0,0 +1,38 @@
+name="megabasterd-bin"
+gives="megabasterd"
+pkgver="8.21"
+pkgdesc="Yet another unofficial (and ugly) cross-platform MEGA downloader/uploader/streaming suite"
+homepage="https://github.com/tonikelope/${gives}"
+depends=("java11-runtime")
```
Elsie19, line 6: Does it only work on java11?

Elsie19 (changes requested):

packages/megabasterd-bin/megabasterd-bin.pacscript
vigress8: Java11 or later. But some of the GitHub runners were pulling Java8 to satisfy the `java-runtime` dependency

Elsie19 (approved):

PR #5081, opened 2023-12-23
fix(obs-backgroundremoval-git): jammy not supported for building anymore

Zahrun:


Zahrun:
CI fails to uninstall

Elsie19 (approved):

PR #5086, opened 2023-12-23
fix(librewolf-app): check if `~/.librewolf` exists before attempting to delete it

vigress8:
Also use `homedir` instead of `~`

### DIFF
--- a/packages/librewolf-app/librewolf-app.pacscript
+++ b/packages/librewolf-app/librewolf-app.pacscript
@@ -34,7 +34,6 @@ post_remove() {
     if ask "Do you want to delete ~/.librewolf/ files?" Y; then
       rm -rf "${homedir}/.librewolf"
     fi
-  else :
   fi
 }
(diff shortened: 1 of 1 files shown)


packages/librewolf-app/librewolf-app.pacscript
```diff
@@ -29,9 +29,12 @@ package() {
...
 post_remove() {
-  # Delete .librewolf
-  if ask "Do you want to delete ~/.librewolf/ files?" Y; then
-    rm -rf ~/.librewolf/
+  # Delete ~/.librewolf if it exists
+  if [[ -d "${homedir}/.librewolf" ]]; then
+    if ask "Do you want to delete ~/.librewolf/ files?" Y; then
+      rm -rf "${homedir}/.librewolf"
+    fi
+  else :
```
Elsie19, line 37: You don't need this line.

Elsie19 (changes requested):

oklopfer (approved):

PR #5129, opened 2023-12-26
upd(emacs): `28.2` -> `29.1`

vigress8:


Elsie19 (changes requested):
Fails on everything except 2 Ubuntu runs.

vigress8:
Dependency naming shenanigans again. Will look into it when I have time

Elsie19 (changes requested):
What are we doing here with removing the config variables?

vigress8:
That level of configuration makes more sense for `emacs-git`. Most users who type `pacstall -I emacs` most likely want to just install the latest stable release of Emacs, without all of that.
It also makes this script an unmaintainable clusterfuck, with dependencies scattered all over the place

PR #5132, opened 2023-12-26
add: `exercism-cli-bin`

raindev:
Tested both on an x86_64 laptop and on an amd64 Pi 4.

### DIFF
--- a/packagelist
+++ b/packagelist
@@ -101,6 +101,7 @@ enpass-deb
 epson-inkjet-printer-stylus-nx110-series-deb
 epson-printer-utility-deb
 exa-git
+exercism
 exploitdb-git
 expressvpn-deb
 eza-git
--- a/packagelist
+++ b/packagelist
@@ -101,7 +101,7 @@ enpass-deb
 epson-inkjet-printer-stylus-nx110-series-deb
 epson-printer-utility-deb
 exa-git
-exercism
+exercism-cli-bin
 exploitdb-git
 expressvpn-deb
 eza-git
--- a/packages/exercism-cli/exercism-cli.pacscript
+++ b/packages/exercism-cli/exercism-cli.pacscript
@@ -1,0 +1,21 @@
+name="exercism-cli"
+repology=("project: exercism-cli")
+pkgver="3.2.0"
+arch=('amd64' 'arm64')
+if [[ ${CARCH} == amd64 ]]; then
+    hash="4ea3e1ea8916a8003da95dbd6eef7a3a29802e637ed6a0f2aaaa2f1c98754915"
+    arch_suffix="x86_64"
+else
+    hash="57f9620e5346bf045ca72b581cb106f9dc56841ac1ff0a3070188c49d5d0b573"
+    arch_suffix="arm64"
+fi
+url="https://github.com/exercism/cli/releases/download/v${pkgver}/exercism-${pkgver}-linux-${arch_suffix}.tar.gz"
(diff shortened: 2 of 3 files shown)


packages/exercism-cli/exercism-cli.pacscript
```diff
@@ -0,0 +1,21 @@
...
+arch=('amd64' 'arm64')
+if [[ ${CARCH} == amd64 ]]; then
+    hash="4ea3e1ea8916a8003da95dbd6eef7a3a29802e637ed6a0f2aaaa2f1c98754915"
+    arch_suffix="x86_64"
+else
+    hash="57f9620e5346bf045ca72b581cb106f9dc56841ac1ff0a3070188c49d5d0b573"
+    arch_suffix="arm64"
+fi
+url="https://github.com/exercism/cli/releases/download/v${pkgver}/exercism-${pkgver}-linux-${arch_suffix}.tar.gz"
+pkgdesc=" A Go based command line tool for exercism.org"
```
Elsie19, line 13:
```suggestion
pkgdesc="Go based command line tool for exercism.org"
```

Elsie19 (changes requested):
Also this should be `exercism-cli-bin`

raindev:
I wasn't sure whether to call the package exercism-cli (more precise) or just exercism (more concise).

Elsie19:
well you are installing a binary, so it should be either `exercism-bin` or `exercism-cli-bin` if there is a gui alternative.

raindev:
@Elsie19 by the same extent the watchexec package should also have been named with -bin suffix, I take?

Elsie19:
> @Elsie19 by the same extent the watchexec package should also have been named with -bin suffix, I take?

👀 actually yeah can you make a PR to fix that?

raindev:
> well you are installing a binary, so it should be either exercism-bin or exercism-cli-bin if there is a gui alternative.

There's no GUI ([used to be](https://github.com/exercism/gui)) alternative but the exercism website itself is also open source.

raindev:
@Elsie19 I'm sorry for missing this in the previous PR, fixed in https://github.com/pacstall/pacstall-programs/pull/5133

Elsie19:
> @Elsie19 I'm sorry for missing this in the previous PR, fixed in https://github.com/pacstall/pacstall-programs/pull/5133

it's ok it was my fault 😳

packagelist
```diff
@@ -101,6 +101,7 @@ enpass-deb
 epson-inkjet-printer-stylus-nx110-series-deb
 epson-printer-utility-deb
 exa-git
+exercism
```
Elsie19, line 104:
```suggestion
exercism-cli-bin
```

Elsie19 (changes requested):

Elsie19 (approved):

packagelist
```diff
@@ -101,7 +101,7 @@ enpass-deb
 epson-inkjet-printer-stylus-nx110-series-deb
 epson-printer-utility-deb
 exa-git
-exercism
+exercism-cli-bin
```
raindev, line 104: That was sloppy of me, thanks for fixing!

raindev: It's quite likely I'll keep forgetting packagelist so I added a commit hook that automates this in https://github.com/pacstall/pacstall-programs/pull/5139

Elsie19: Thanks!

PR #5133, opened 2023-12-26
fix(watchexec): update the name to `watchexec-bin`

raindev:
Given the package is actually binary

### DIFF
--- a/packagelist
+++ b/packagelist
@@ -461,7 +461,7 @@ waf
 walc-app
 warp
 warpinator-deb
-watchexec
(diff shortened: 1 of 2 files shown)


Elsie19 (changes requested):
Fix `packagelist` as well.

raindev:
I hope no one have installed the new package in the meantime. If someone did that results in a conflict and the old package needs to be removed manually (`apt remove watchexec`).

Elsie19 (approved):

PR #5139, opened 2023-12-28
feat: add a pre-commit hook to update packagelist

raindev:
Replace the check that verifies the order with the one that actually updates the file with all entries from `packages/`.


raindev:
I have inlined script - makes sense to avoid a new file for a single command.

raindev:
@Elsie19 what do you think?

Elsie19 (approved):

PR #5142, opened 2023-12-29
fix(aseprite): use correct patch filename

vigress8:
In preparation for pacstall/pacstall#1020

### DIFF
--- a/packages/aseprite/aseprite.pacscript
+++ b/packages/aseprite/aseprite.pacscript
@@ -27,7 +27,7 @@ prepare() {
   unzip -qq "Skia-Linux-Release-x64-libc++.zip" -d skia
 
   # Apply patch from AUR
-  patch -stp1 <PACSTALL_patchesdir/"desktop.patch?h=aseprite"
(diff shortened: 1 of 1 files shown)


packages/aseprite/aseprite.pacscript
```diff
@@ -27,7 +27,7 @@ prepare() {
   unzip -qq "Skia-Linux-Release-x64-libc++.zip" -d skia
 
   # Apply patch from AUR
-  patch -stp1 <PACSTALL_patchesdir/"desktop.patch?h=aseprite"
+  patch -stp1 <PACSTALL_patchesdir/desktop.patch
```
oklopfer, line 30: Turns out this doesn't work as the pacstall PR had intended. 

Now, curl -sO is split between two versions:
on jammy and below, it still produces `desktop.patch?h=aseprite`
on bookworm and above, it produces `desktop.patch`

Your pacscript is now going to have to figure out how to conform to this.

oklopfer (changes requested):

PR #5150, opened 2024-01-01
upd(hyprland): `0.33.1` -> `0.34.0`

Elsie19:


oklopfer (changes requested):
Nope. Don’t add and upd at the same time.

PR #5217, opened 2024-01-17
upd(thorium-deb): `117.0.5938.157` -> `119.0.6045.214`

vigress8:


packages/thorium-deb/thorium-deb.pacscript
```diff
@@ -1,14 +1,31 @@
...
+  arm64)
+    url="https://github.com/Alex313031/Thorium-Raspi/releases/download/M${pkgver}/${gives}_${pkgver}_${CARCH}.deb"
+    hash="811dded7557b0ca9a93c3a50e5ce6658f2d185b9858443a76d359485598856b5"
+    ;;
+  i386)
+    url="https://github.com/Alex313031/Thorium-Other/releases/download/${pkgver}-2/${gives}_${pkgver}_${CARCH}.deb"
+    hash="ee5443c8547bb1828e7a105fb5f1ca4eaa43d5a16fd683a37bb1b2e9f57db21c"
+    ;;
+  amd64)
+    if grep -q avx2 /proc/cpuinfo; then
```
Elsie19, line 16: What's avx?

Elsie19 (changes requested):

packages/thorium-deb/thorium-deb.pacscript
vigress8: A CPU feature, google it. Thorium and Mercury provide specialized builds with it

Elsie19: I don't like grepping in a pacscript, I *really* want all pacscripts to be simple control flow with no external commands.

Elsie19 (changes requested):

packages/thorium-deb/thorium-deb.pacscript
vigress8: Again, if you have a better way I'm all ears. Pacstall does not have a way to specify compatible hardware features

Elsie19: Then we should make one! Should we discuss on discord/matrix?

oklopfer, line 16: The idea with https://github.com/pacstall/pacstall/pull/1025 is you should be able to replace 
```bash
if grep -q avx2 /proc/cpuinfo; then
```
with
```bash
if [[ " ${DCPUFLAGS[*]} " =~ " avx2 " ]]; then
```

PR #5218, opened 2024-01-17
add: `mercury-browser-deb`

vigress8:


packages/mercury-browser-deb/mercury-browser-deb.pacscript
```diff
@@ -0,0 +1,14 @@
+name="mercury-browser-deb"
+gives="mercury-browser"
+pkgver="121.0.1"
+arch=('amd64')
+if grep -q avx /proc/cpuinfo; then
```
Elsie19, line 5: what's avx?

Elsie19 (changes requested):

packages/mercury-browser-deb/mercury-browser-deb.pacscript
Zahrun: A highlight of these optimizations, and the first we will discuss, is [AVX](https://en.wikipedia.org/wiki/Advanced_Vector_Extensions). AVX, or Advanced Vector Extensions, is an [instruction set](https://en.wikipedia.org/wiki/Instruction_set_architecture) extension for 2nd Gen Core (Intel), and AMD FX Processors and newer. It enables contracting mathematical expressions in source code into a vector format in which the CPU can perform multiple operations on data at a time, so it is a [SIMD](https://en.wikipedia.org/wiki/Single_instruction,_multiple_data) instruction set like its [predecessors](https://en.wikipedia.org/wiki/Streaming_SIMD_Extensions#Later_versions): MMX, SSE, SSE2, SSE3, and SSE4. The Chromium Project makes heavy use of code that can be vectorized, and so AVX is a logical step in improving performance. The only reason Chromium does not use it by default is compatibility: older processors (pre-2011) dont have AVX capability, and so cannot run programs compiled with AVX. To combat this, and to provide Thorium to a wider audience, I occassionally make releases for Linux and Windows that are SSE4/SSE3 only.

https://thorium.rocks/optimizations

also applies to mercury browser

PR #5484, opened 2024-01-21
fix(aseprite): adjust to proper `patch` filename

oklopfer:


oklopfer:
oops I didnt see https://github.com/pacstall/pacstall-programs/pull/5142

PR #5546, opened 2024-01-25
fix(fastfetch-git): deleted libmagickcore-dev dependency (called twice)

Xdavius:
not critical: 
 libmagickcore-dev was called twice as makedepends


Elsie19 (approved):
Ha thanks for the catch!

PR #5548, opened 2024-01-25
fix(hyprland.pacscript): libdrm from pacdepends to makedepends

Xdavius:
libdrm from pacdepends to makedepends


oklopfer:
We created those packages specifically for pacstall because their dependencies don’t work on hyprland. Closing.

Xdavius:
> We created those packages specifically for pacstall because their regular dependencies don’t work on hyprland. Closing.

those package make falling wine-staging-git ...

it's a mess :(

PR #5549, opened 2024-01-25
Fix(xdg-desktop-portal-hyprland): Moved libdrm from pacdeps to make depends

Xdavius:
…pends


Xdavius:
libdrm package not needed here

hyperland is not compaible with :  

'ubuntu:devel' 'debian:sid' 'ubuntu:mantic' 'ubuntu:noble')

PR #5550, opened 2024-01-25
upd(libdrm): `2.4.119` -> `2.4.120`

Xdavius:
resolve dependencies loop

libdrm-dev/testing,unstable 2.4.120-1 amd64


Xdavius:
semms this is the root of all my dependencies loops

oklopfer:
> semms this is the root of all my dependencies loops

Ah. The provides libdrm2 makes sense. Will look into merging later today. Also, can you please make sure the title is all lowercase for the merge request. Thanks

Xdavius:
Yes. Tomorrow (France here)
I need to investigate a little bit more, and solve the problem because all of this makes wine-staging-git.pacstall conflicting is a way I don't understand.

Still in discussion with Antiz from archlinux staff about this.

oklopfer (approved):

PR #5566, opened 2024-01-28
fix(stockfish-git): improve build sequence

pythoncrazy:
I missed the key step of stripping the exectuable to make it smaller and faster, as well as using profile-build to do something good. This has been tested on my jetson xavier and jetson nano, so it will probably work. Thank y'all!

### DIFF
--- a/packages/stockfish-git/stockfish-git.pacscript
+++ b/packages/stockfish-git/stockfish-git.pacscript
@@ -13,7 +13,6 @@ pkgver() {
 build(){
   cd src
   make -j"${NCPU}" profile-build
-
 }
 
 package() {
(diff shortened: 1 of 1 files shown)


Elsie19:
Fix the title.

pythoncrazy:
Title should be better

packages/stockfish-git/stockfish-git.pacscript
```diff
@@ -12,10 +12,12 @@ pkgver() {
 
 build(){
   cd src
-  make -j"${NCPU}" build
+  make -j"${NCPU}" profile-build
+
```
Elsie19, line 16:
```suggestion
```

Elsie19 (changes requested):

Elsie19 (approved):

PR #5570, opened 2024-01-29
fix(steam-deb): doesn't setup architecture in prepare.

Xdavius:
changed prepare() to pre_install()


Xdavius:
I'm not able to improve this package.
Tried to install it from fresh Debian Sid today but needed to add-architecture manualy because of an error.

Nevermind, we all have the same Steam because it is auto updated.

I just wanted to avoid a step for users

PR #5584, opened 2024-02-01
add: `katawa-shoujo-re-engineered`

vigress8:


vigress8:
#5585 should make this work.

PR #5610, opened 2024-02-07
fix(*.pacscript): remove useless `vim:set` lines

vigress8:
The `shfmt` hook exists already, and people can just add this to their personal config if they desire.


vigress8:
Just add a vimrc sample for `*.pacscript` to the wiki. In any case it doesn't make sense to have this line in some files but not others

PR #5614, opened 2024-02-08
fix(glow-deb): fix arch64 is not a valid architecture

lfromanini:
Can't install `glow-deb` with `pacstall -I glow-deb` due to `[!] ERROR: 'arch64' is not a valid architecture`.


Elsie19:
Don't worry it's just debian Sid being stupid. I'll go ahead and merge this.

Elsie19 (approved):

PR #5641, opened 2024-02-18
upd(oil): `0.17` -> `0.20`

srd424:
https://www.oilshell.org/ is still unstable and making changes, so it's good to be able to follow the latest version to keep up.

Not sure of etiquette here - @wizard-28 is marked is maintainer but doesn't seem to have been around for a bit? Will mark this as a draft for a while in the hope of getting some feedback on what's best to do.

I've deployed this locally and only needed one single-line patch to one of my scripts due to an apparent syntax change - to be expected as this is explicitly an unstable package. Other than that, nothing has exploded, and upstream has a pretty heavy test suite anyway.

### DIFF
--- a/packages/oil/oil.pacscript
+++ b/packages/oil/oil.pacscript
@@ -7,12 +7,12 @@
 maintainer="wizard-28 <wiz28@pm.me>"
 
 name="oil"
-pkgver="0.17.0"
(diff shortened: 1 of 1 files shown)


Elsie19:
Fix title.

Elsie19 (approved):

PR #5681, opened 2024-03-01
upd(git-butler-deb): `0.10.6-619` -> `0.10.19-680`

nezred:


nezred:
Is there an issue with the failed tests? they fail because there isnt a source attribute, which i dont understand why it happens.

Elsie19:
I think I said this elsewhere but it's because of a breaking commit in develop.

Elsie19 (approved):

PR #5697, opened 2024-03-04
upd(floorp-bin): `11.8.2` -> `11.10.5`

Zahrun:


Elsie19 (changes requested):
Spacing is off.

Zahrun:
Oh also I had not seen the part about arm64. I ran pacup and it didn't detect the two hashes

Elsie19 (approved):

PR #5714, opened 2024-03-05
add: `neovide-freedesktop`

nezred:
Since i need to get the desktop icons from the git repo, and i cant put multiple sources in the neovide-bin repo, so im just installing the desktop files through here.

I think we could put this as dependency to that package, and that way we can allow that package to include the app menu entry and icons.

Although i dont know if it isnt worth it, since it requires the whole repo to be downloaded just to get that part (unless there is a way to download a folder in a github repo through a github url)

### DIFF
--- a/packagelist
+++ b/packagelist
@@ -280,6 +280,7 @@ nano
 nautilus-open-in-blackbox-git
 neofetch
 neovide-bin
+neovide-freedesktop
 neovim
 neovim-app
 neovim-git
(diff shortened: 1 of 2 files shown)


nezred (commented):
also, should this package be named something else? i thought this was appropriate, but maybe there is a better name idk

Elsie19:
This should be added to the neovide package.

nezred:
do i add the pacdep for neovide-freedesktop in neovide as a different PR, or here?

Elsie19:
Look at how other packages like `amfora-bin` handle desktop files.

nezred:
ah okay thanks, i hadnt seen this. Should I put wget as a make dependency? since it is needed for downloading the files, or is it a dependency of pacstall?

Elsie19:
It's already a dep on pacstall.

nezred:
I've made the PR for neovide-bin. Thanks for the help Elsie 😄

PR #5725, opened 2024-03-08
fix(discord-canary): prevent force update message

Elsie19:


vigress8:
Does Discord create this file at launch? If so, this needs testing to make sure nothing breaks

Elsie19:
I don't believe so but let me check.

Elsie19:
Have confirmed it does not create on the first go.

vigress8 (approved):
LGTM, might as well add this for `discord` too

PR #5752, opened 2024-03-18
add: `emacs-snapshot-deb`

vigress8:


vigress8:
@Elsie19 this won't work until it's merged, so take a look

PR #5756, opened 2024-03-20
refactor(*): move `ln` functions to `package()`

oklopfer:


vigress8 (changes requested):
The `-f` flag can be removed from all of the `ln` invocations, since nothing is supposed to be overwritten.

PR #5762, opened 2024-03-22
add: `ripgrep-deb`

jerengie:
This PR adds a package for ripgrep. The `.deb` comes from the upstream repo of `ripgrep`.


Elsie19:
Is ripgrep in the apt repos? If so, what's the version on Jammy?

jerengie:
> Is ripgrep in the apt repos? If so, what's the version on Jammy?

Yes, it's in the repos. The version on Jammy is 13.0.0.

Elsie19:
> > Is ripgrep in the apt repos? If so, what's the version on Jammy?
> 
> Yes, it's in the repos. The version on Jammy is 13.0.0.

Ok because here's my problem: I don't want packages that exist at a reasonable version in the repos to be added to pacstall. Basically I don't want pacstall to serve as an updated Ubuntu repo.

vigress8:
> Ok because here's my problem: I don't want packages that exist at a reasonable version in the repos to be added to pacstall. Basically I don't want pacstall to serve as an updated Ubuntu repo.

Why?

Elsie19:
> > Ok because here's my problem: I don't want packages that exist at a reasonable version in the repos to be added to pacstall. Basically I don't want pacstall to serve as an updated Ubuntu repo.
> 
> Why?

Sorry what I meant is an updated Ubuntu repo where it's full of packages that are slightly ahead of the Ubuntu repos, if that makes any sense 😆

jerengie:
From my point of view, this is perfectly understandable. Shall I close this PR then?

PR #5808, opened 2024-03-30
add: `stremio-deb`

Rockstar9604:
Need #5807 and #5806 to work properly

### DIFF
--- a/packagelist
+++ b/packagelist
@@ -426,6 +426,7 @@ steam-launcher-deb
 steamtinkerlaunch-git
 stellarium
 stockfish-git
+stremio-deb
 sublime-merge-deb
 sublime-text-deb
 surf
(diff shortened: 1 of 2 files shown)


Elsie19 (changes requested):
This only works on one distro, and you don't have `incompatible`, so there is a problem here.

Rockstar9604:
> This only works on one distro, and you don't have `incompatible`, so there is a problem here.

Oh… I had tested it on PopOS, and I didn’t encounter any issues except for the two missing libraries. I tried to find a solution, but I don't find nothing usefull.

Elsie19:
Add all the distros that failed in the CI to `incompatible`.

Elsie19 (approved):

PR #5818, opened 2024-04-01
fix(lazarus-deb): Changed to lazarus-project-deb

Xdavius:
Upstream name is different to Debian name, and cannot use provide as the .deb got his own provide. So the dependency need is cover by other apps

This fix the problem between pacstall db and apt db so pacstall -R can remove the .deb


Xdavius:
After this is merged, I will push goverlay. This change was needed for it to work

Elsie19 (approved):

PR #5824, opened 2024-04-03
upd(goverlay): `1.0` -> `1.1.1`

Xdavius:


Xdavius:
I have canceled my branch.. it has canceled all my PR ...

Xdavius:
Here the control file generated : 

```Depends: mangohud, fpc-laz, fpc-src, lazarus-project, libqt6pas6
Package: goverlay
Version: 1.1.1-pacstall1
Architecture: amd64
Section: Pacstall
Priority: optional
Build-Depends: git, libglu1-mesa-dev, qt6-base-dev
Conflicts: goverlay-bin
Replace: goverlay-bin
Maintainer: Xdavius <xdavius@github.com>
Description: A GUI to help manage Vulkan/OpenGL overlays
Installed-Size: 4193
```

Xdavius:
Hi. Thank you for replying.

In fact, I have tested on my side. Uninstalling fpc-src uninstall goverlay. But I don't want this comportement... Because it's not an execution dependency, just a build one.

If you know the way to avoid this problem... (Not tested to declare empty depends, I don't know if it will do the trick...)

For your question, goverlay-bin is the binary provided by upstream. At this moment, it is still in qt5 version 1.0.

This one is qt6 and rebuilt from source. The distributions aren't totally compatible with it, making the use of qt6ct for exemple for the theme.

Originally, goverlay-bin was named goverlay, but was a bin. So I have updated the name of the package.

And goverlay is from source as it must be.

I will continue to maintain both, but -bin will be left on 1.0 version for a while. 

Waiting for Noble to see If I can update without problems.

Xdavius:
Hi. Thank you for replying.

In fact, I have tested on my side. Uninstalling fpc-src uninstall goverlay. But I don't want this comportement... Because it's not an execution dependency, just a build one.

If you know the way to avoid this problem... (Not tested to declare empty depends, I don't know if it will do the trick...)

For your question, goverlay-bin is the binary provided by upstream. At this moment, it is still in qt5 version 1.0.

This one is qt6 and rebuilt from source. The distributions aren't totally compatible with it, making the use of qt6ct for exemple for the theme.

Originally, goverlay-bin was named goverlay, but was a bin. So I have updated the name of the package.

And goverlay is from source as it must be.

I will continue to maintain both, but -bin will be left on 1.0 version for a while. 

Waiting for Noble to see If I can update without problems.

Xdavius:
@Elsie19 I think it's good now.

packages/goverlay/goverlay.pacscript
```diff
@@ -1,26 +1,37 @@
...
-depends=("mangohud" "libqt5pas-dev")
-pkgdesc="Graphical UI to help manage Linux overlays"
-maintainer="Elsie19 <elsie19@pm.me>"
+replace=("goverlay-bin")
+pkgver="1.1.1"
+pkgdesc="A GUI to help manage Vulkan/OpenGL overlays"
+arch=("amd64")
+makedepends=('git' 'libglu1-mesa-dev' 'qt6-base-dev')
+case "${DISTRO}" in
+debian:testing|debian:sid)
```
Elsie19, line 10:
```suggestion
debian:testing | debian:sid)
```

packages/goverlay/goverlay.pacscript
```diff
@@ -1,26 +1,37 @@
...
+  pacdeps=('mangohud' 'fpc-laz-deb' 'fpc-src-deb' 'lazarus-project-deb' 'qt6pas')
+  ;;
+esac
+url="https://github.com/benjamimgois/goverlay/archive/refs/tags/${pkgver}.tar.gz"
+hash="06dcad9ab7db990f2df04be01d549953e274907d7f691fe1bc4bbe712329936c"
 repology=("project: goverlay")
 
+build() {
+  case "${DISTRO}" in
+  debian:testing|debian:sid)
```
Elsie19, line 24:
```suggestion
  debian:testing | debian:sid)
```

Elsie19 (changes requested):

Xdavius:
@oklopfer possible to merge ? or miss something ?

packages/goverlay/goverlay.pacscript
```diff
@@ -1,26 +1,37 @@
...
-Exec=goverlay
-Icon=goverlay
-Terminal=false
-Type=Application
-Categories=Game;
-Keywords=MangoHud;vkBasalt;ReplaySorcery;msi;afterburner;' | sudo tee "${pkgdir}/usr/share/applications/goverlay.desktop" > /dev/null
-  sudo wget -q https://raw.githubusercontent.com/benjamimgois/goverlay/master/data/icons/128x128/goverlay.png -P "${pkgdir}/usr/share/icons/hicolor/128x128/apps/"
+  sudo make prefix=/usr libexecdir=/lib DESTDIR="${pkgdir}" install
+  # Fix Issue to change Depends with usage of pacdeps
+  sudo sed -i "/^Depends:*/c\Depends: mangohud, libqt6pas-dev" "${pkgdir}"/DEBIAN/control
```
Elsie19, line 36: @Xdavius we're still deliberating on this line. That's why this hasn't been merged yet.

Xdavius: Can be changed to prepare and install in package. You are right.

I will change it.

Xdavius: @Elsie19 here the build of my friend Antiz for arch : 
https://gitlab.archlinux.org/archlinux/packaging/packages/goverlay/-/blob/main/PKGBUILD?ref_type=heads

And for my workaroud yes... I understand the problem.

We need a new section names `makepacdeps` to use pacstall as make dependency and not runtime dep.

For now, I suppose there is no other solution by editing the Debian Control file in this particular case.

On my repository, I have made worst because of missing pre_instal function :/ 

When pacstall will be ready for a new syntaxe, we will update the package and fix the workaround.

Other thing, I don't about pacstall 5.0 and the changes if it will be compatible... Well, if it break, we will patch XD

D-Brox:
This needs to be updated to the 5.0.0 spec

Xdavius:
> This needs to be updated to the 5.0.0 spec

use goverlay-bin until I do the changes here. Have to test with v5.

Elsie19:
There are merge conflicts now.

packages/goverlay/goverlay.pacscript
```diff
@@ -1,33 +1,46 @@
...
+  makedepends+=("lazarus" "fpc" "fpc-source")
+  ;;
+*)
+  pacdeps=('mangohud' 'fpc-laz-deb' 'fpc-src-deb' 'lazarus-project-deb' 'qt6pas')
+  ;;
+esac
+
+
+build() {
+  export external_connection=true
```
oklopfer, line 26: Don’t export this. It should be a pacscript variable outside of the functions.

oklopfer (changes requested):

Elsie19 (approved):

Issue #5873, opened 2024-04-16
PacReq: EnvyControl


Chromiell:
### Types

-deb (Debian achieve release of the package)

### URL

The package can be found at this GitHub repo, it already comes precompiled in a .deb format: https://github.com/bayasdev/envycontrol/releases

### Building guide

https://github.com/bayasdev/envycontrol

### Additional context

EnvyControl is a neat little Python script that allows the user to switch between Integrated, Hybrid and Dedicated graphics modes for Nvidia laptops, it's very simple to use, light and doesn't really require any additional configuration. It can be easily installed as a .deb package directly from the Release page on their GitHub page, but it would be great to have it as a Pacstall.

Xdavius:
This package is in stand by for the moment by the dev. Also it doesn't support Wayland.

I'm waiting for the integration by the dev.

PR #5880, opened 2024-04-17
add: `envycontrol-deb`

ItsLynix:


ItsLynix:
solves #5873

ItsLynix:
Will follow up on this one

PR #5888, opened 2024-04-21
fix(rhino*core): repair swapping between each other

oklopfer:
I guess the issue was just `--no-rename` and not also `post_remove`. no idea man.


oklopfer:
ok nope, `post_remove` definitely causing issues still. I hate this.

PR #5935, opened 2024-05-03
refactor(*): post-transition cleanup

vigress8:
remove uses of `gnuarch` in `source`, redundant `post_{install,remove}()` and `sudo` usage


packages/librewolf-app/librewolf-app.pacscript
```diff
@@ -14,23 +14,19 @@ sha256sums_arm64=(
...
-  gnuarch="x86_64"
-else
-  gnuarch="aarch64"
-fi
 source=(
-  "https://gitlab.com/api/v4/projects/24386000/packages/generic/librewolf/${pkgver}/LibreWolf.${gnuarch}.AppImage"
   "icon.png::https://gitlab.com/uploads/-/system/group/avatar/5407097/icon.png"
   "${gives}.desktop::https://raw.githubusercontent.com/pacstall/pacstall-programs/master/packages/${pkgname}/${gives}.desktop"
 )
+source_amd64=("https://gitlab.com/api/v4/projects/24386000/packages/generic/librewolf/${pkgver}/LibreWolf.x86_64.AppImage")
```
oklopfer, line 21: this one needs to deprecate the old source array and just double it up for each CARCH. Remember CARCH will get appended *after* regular var

oklopfer (changes requested):

packages/docker-compose-plugin-bin/docker-compose-plugin-bin.pacscript
```diff
@@ -9,16 +9,12 @@ repology=("project: docker-compose")
...
-source=("https://github.com/docker/compose/releases/download/v${pkgver}/docker-compose-linux-${gnuarch}")
+source_amd64=("https://github.com/docker/compose/releases/download/v${pkgver}/docker-compose-linux-x86_64")
+source_arm64=("https://github.com/docker/compose/releases/download/v${pkgver}/docker-compose-linux-aarch64")
 
 package() {
   cd "${_archive}"
   mkdir -p "${pkgdir}/usr/libexec/docker/cli-plugins"
-  chmod a+x ./"docker-compose-linux-${gnuarch}"
-  mv ./"docker-compose-linux-${gnuarch}" "${pkgdir}/usr/libexec/docker/cli-plugins/docker-compose"
+  chmod a+x ./docker-compose-linux-*
```
oklopfer, line 18: Don't use a glob here. Just move the gnuarch check to inside of the package function here.

oklopfer (changes requested):

packages/librewolf-app/librewolf-app.pacscript
```diff
@@ -14,23 +14,19 @@ sha256sums_arm64=(
...
+source_arm64=("https://gitlab.com/api/v4/projects/24386000/packages/generic/librewolf/${pkgver}/LibreWolf.aarch64.AppImage")
 breaks=("${gives}-deb" "${gives}-bin" "${gives}-git")
 repology=("project: librewolf")
 
 package() {
   cd "${_archive}"
   # Install appimage
-  install -Dm755 "LibreWolf.${gnuarch}.AppImage" "${pkgdir}/usr/bin/${gives}"
+  install -Dm755 LibreWolf.*.AppImage "${pkgdir}/usr/bin/${gives}"
```
oklopfer, line 29: Same issue, don't glob, just move the check to inside of the function for this part

oklopfer (changes requested):

packages/librewolf-app/librewolf-app.pacscript
vigress8: Can't the order of the sums be switched instead?

oklopfer: logistically speaking, yes that would work. But for maintainability, the primary source should always come first.

PR #5942, opened 2024-05-04
fix(*.pacscript): convert rest of `homepage` -> `url`

Elsie19:


D-Brox (approved):
I assume that's all the cases

PR #5991, opened 2024-05-07
upd(wine-staging-git): `8.13` -> `9.8`

Xdavius:
Still testing and need tests for depends and break (wine-fonts). Is it really the staging version ???
please, need to be tested, and source need to be confirmed.

do not merge. thanks

translated from https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=wine-git


Xdavius:
Love dependency hell...
So, in fact, this package wasn't staging but only upstream.
will change all to staging if I can

https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=wine-staging-git

Xdavius:
```satisfy:command-line : Depends: oss4-dev but it is not installable or
                                 alsa-oss but it is not installable```

Nothing to do here. We have to wait for oss4-dev on testing. 

Good for me.

oklopfer (approved):

PR #5999, opened 2024-05-10
upd(virtualbox-deb): `7.0.14` -> `7.0.18`

Xdavius:


Elsie19:
To be clear, you want this to *not* work on anything except debian stable?

Xdavius:
> To be clear, you want this to _not_ work on anything except debian stable?

Cannot make miracles. Debs are ony for Ubuntu 22.04 or under and Debian stable or under... So the pacscript can only support now thoses distro or under but they will disapears and aren't on pacstall checks anymore... 

when Oracle will support Ubuntu 24.04, I will update the package. ATW, still no official deb from them... 

And impossible to install with a trick virtualbox deb from sid for testing or sid because debs are splited...

It will be a great feature to support installing multiple -debs ...

Xdavius:
@Elsie19 Later, I can write virtualbox-bin pacscript with the .run source https://download.virtualbox.org/virtualbox/7.0.16/VirtualBox-7.0.16-162802-Linux_amd64.run

I have to study this, but for debs, we are good. Waiting for you =)

Elsie19 (approved):

PR #6006, opened 2024-05-12
feat: add .shellcheckrc, specify shell=bash, re-enable codes which are safe to enable

vigress8:
Rationale: people constantly get confused when running shellcheck manually. This makes those invocations consistent with the pre-commit hook (which the args can be removed from now). Also suppresses redundant LSP warnings.

3 of the 4 codes I re-enabled triggered no warnings when run on all files, the fourth - only for one file, which I fixed.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -18,7 +18,6 @@ repos:
     hooks:
       - id: shellcheck
         types: []
-        args: ["-o", "all", "-e", "2148", "-e", "SC2033", "-e", "SC2032", "-e", "SC2034", "-e", "SC2154", "-e", "SC2312", "-e", "SC2164", "-e", "SC2143", "-e", "SC2103"]
         files: '^.*\.(pacscript|sh)$'
   - repo: https://github.com/maxwinterstein/shfmt-py
     rev: v3.7.0.1
--- a/.shellcheckrc
+++ b/.shellcheckrc
@@ -1,0 +1,22 @@
+enable=all
+shell=bash
+
+# foo appears unused. Verify it or export it.
+# https://www.shellcheck.net/wiki/SC2034
+disable=SC2034
+
+# Use a ( subshell ) to avoid having to cd back.
+# https://www.shellcheck.net/wiki/SC2103
+disable=SC2103
+
+# var is referenced but not assigned.
+# https://www.shellcheck.net/wiki/SC2154
+disable=SC2154
+
+# Use cd ... || exit in case cd fails.
+# https://www.shellcheck.net/wiki/SC2164
+disable=SC2164
+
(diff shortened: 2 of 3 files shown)


.pre-commit-config.yaml
```diff
@@ -18,7 +18,6 @@ repos:
     hooks:
       - id: shellcheck
         types: []
-        args: ["-o", "all", "-e", "2148", "-e", "SC2033", "-e", "SC2032", "-e", "SC2034", "-e", "SC2154", "-e", "SC2312", "-e", "SC2164", "-e", "SC2143", "-e", "SC2103"]
```
oklopfer, line 21: why are only some of the exceptions included in the new file?

vigress8: Read the comment

oklopfer: It doesn't explain why you removed half of them

vigress8: > 3 of the 4 codes I re-enabled triggered no warnings when run on all files, the fourth - only for one file, which I fixed.

Disabling those codes is not relevant anymore

oklopfer: why?

vigress8: Because they trigger no warnings when run on all pacscripts, therefore, whatever the reason for disabling them was, there isn't one anymore

oklopfer: Future proofing? Reading them, they're clearly not unimportant:
https://www.shellcheck.net/wiki/SC2148

oklopfer: https://www.shellcheck.net/wiki/SC2033
These are shellchecks that are disabled so they don't fail on pacscripts and how they're structured.

oklopfer: https://www.shellcheck.net/wiki/SC2143
This one is a styling choice.

oklopfer: As is https://www.shellcheck.net/wiki/SC2103

vigress8: Ok so, SC2148 is made irrelevant by explicitly specifying shell=bash, because all pacscripts are in bash. SC2103 is irrelevant because we had to use `cd ..` a lot before, but don't anymore. The other two I just think are fine to re-enable, considering there's no good reason for disabling them

oklopfer: for 2103, we want to discourage the usage of subshells, so disabling it is important imo

vigress8: Yeah but one can use pushd / popd instead of cd in those cases

oklopfer: Just re-disable this one please. I'm fine with ceding the others

oklopfer (approved):

PR #6018, opened 2024-05-16
upd(neovim): `0.9.5` -> `0.10.0`

Elsie19:


Elsie19:
I accidentally pushed `mold -run` in the PR because I was tired of bad compile times when testing something.

oklopfer (approved):

PR #6027, opened 2024-05-18
fix(hamclock-big): fix the desktop shortcut

fang64:
I made a mistake in the desktop shortcut, so this is to fix that before adding the source file for the next PR.


Elsie19:
Bump pkgrel as well.

fang64:
Ok, I've done that.

oklopfer (approved):

PR #6029, opened 2024-05-19
fix(hamclock-big): use and source desktop shortcut

fang64:


packages/hamclock-big/hamclock-big.pacscript
```diff
@@ -1,13 +1,19 @@
 pkgname="hamclock-big"
 pkgver="4.00"
-pkgrel="2"
-source=("https://github.com/kj7rrv/hamclock/archive/refs/tags/v${pkgver}.tar.gz")
+pkgrel="3"
+source=(
+  "https://github.com/kj7rrv/hamclock/archive/refs/tags/v${pkgver}.tar.gz"
+  "hamclock.desktop::https://raw.githubusercontent.com/pacstall/pacstall-programs/master/packages/${pkgname}/hamclock.desktop"
```
oklopfer, line 6: you don't need the link here. you can just use "hamclock.desktop"

oklopfer (changes requested):

packages/hamclock-big/hamclock-big.pacscript
fang64: Alright I'll make the change, so files without a url will pull from the repository. Good to know!

oklopfer (approved):

PR #6051, opened 2024-05-23
add: `hamclock-bigger`

fang64:


oklopfer:
I'm almost positive that if you add the .desktop file to the source list now, it should still work. Relative files match the repo of their parent, and since tests are run on the repo of the PR (and not the origin repo), it should still be able to pick it up just fine.

fang64:
> I'm almost positive that if you add the .desktop file to the source list now, it should still work. Relative files match the repo of their parent, and since tests are run on the repo of the PR (and not the origin repo), it should still be able to pick it up just fine.

I did try to run it, but I'll give it a shot again it.

oklopfer:
Looks like it did :) the trick is the CI uses the `-Qa` command to switch the repositories temporarily. So we can make it work as long as the relative file has been uploaded to the same branch :)

oklopfer (approved):

PR #6053, opened 2024-05-24
feat(*-app): disable appimagelauncher in desktop file

Elsie19:


oklopfer:
so what it do exactly?

Elsie19:
> so what it do exactly?

So anytime someone has `appimagelauncher` installed, it will hook into the execution of any appimage and will prompt a message looking like this:

image

With this PR, we explicitly tell appimagelauncher to *not* hook into our installed package, because we have already setup the desktop file and everything.

oklopfer (approved):

PR #6077, opened 2024-06-03
fix(lazygit): install path

sarmong:


packages/lazygit/lazygit.pacscript
```diff
@@ -10,6 +10,6 @@ sha256sums=("ff05cb8ebd0785ce220975e5cd394bc200424d9a58727fa81e0bbb3828db7ff1")
 
 package() {
   cd "${_archive}"
-  install -Dm755 "lazygit" "/usr/local/bin/lazygit"
+  install -Dm755 "lazygit" "${pkgdir}/usr/local/bin/lazygit"
```
vigress8, line 13:
```suggestion
  install -Dm755 "lazygit" "${pkgdir}/usr/bin/lazygit"
```
More standard install path for binaries

sarmong: The binaries are not managed by apt, so, in my opinion, `/usr/local/bin` is a much better location for them. But as you wish.

vigress8: Just to be clear, it's merely a suggestion, you're free to choose what you consider reasonable!

sarmong: I meant it as a general comment regarding pacstall-installed programs. But if all (or most) of them are already being installed in `/usr/bin`, I think consistency is preferable.

Elsie19 (approved):

PR #6093, opened 2024-06-07
fix({fsearch,jellyfin-media-player,veracrypt}-deb): use regular `source` as else arrays

oklopfer:


oklopfer:
should have been fixed by https://github.com/pacstall/pacstall/pull/1131, but seems to have broken since it was merged already. awesome.

PR #6138, opened 2024-06-17
add: `hyfetch-git`

pythoncrazy:
Hi, I have added hyfetch-git, based upon https://github.com/hykilpikonna/hyfetch/ & https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=hyfetch-git

### DIFF
--- a/hyfetch-git/hyfetch-git.pacscript
+++ b/hyfetch-git/hyfetch-git.pacscript
@@ -18,5 +18,5 @@ package() {
   cd "${_archive}"
   chmod +x neofetch
   chmod +x hyfetch/scripts/neowofetch
-   /usr/bin/python3 setup.py install --root="${pkgdir}" --optimize=1
+  /usr/bin/python3 setup.py install --root="${pkgdir}" --optimize=1
 }
--- a/hyfetch-git/hyfetch-git.pacscript
+++ b/hyfetch-git/hyfetch-git.pacscript
@@ -1,0 +1,22 @@
+pkgname="hyfetch-git"
+gives="hyfetch"
+pkgdesc="Neofetch with LGBTQ pride flags."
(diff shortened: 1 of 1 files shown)


hyfetch-git/hyfetch-git.pacscript
```diff
@@ -0,0 +1,22 @@
+pkgname="hyfetch-git"
+gives="hyfetch"
+pkgdesc="Neofetch with LGBTQ pride flags."
```
Elsie19, line 3:
```suggestion
pkgdesc="Neofetch with LGBTQ pride flags"
```

hyfetch-git/hyfetch-git.pacscript
```diff
@@ -0,0 +1,22 @@
...
+makedepends=('python3-pip')
+external_connection="true"
+
+build() {
+  cd "${_archive}"
+  pip install .
+}
+
+package() {
+  cd "${_archive}"
```
Elsie19, line 18: Do not use `${_archive}` in new scripts.

Elsie19 (changes requested):

hyfetch-git/hyfetch-git.pacscript
pythoncrazy: What is the modern alternative? Is it `cd "${srcdir}/${pkgname%-git}"`? Thank you for all of your wonderful help and support!

Elsie19: Well first of all, it should be `${pkgname%-git}`, but I would suggest just not making source download without the git, so then you wouldn't have to have `${pkgname%-git}` everywhere, and just have `${pkgname}`.

oklopfer (approved):

PR #6143, opened 2024-06-19
ci(srcinfo.sh): fix `.SRCINFO` fmting

oklopfer:
this script is damn slow

don't merge yet

required for https://github.com/pacstall/pacstall/pull/1153's full pkgbase usage to be useful

---
`.SRCINFO` generation is now required to be done *before* `packagelist` generation. Both are handled by `./scripts/srcinfo.sh`:
```bash
./scripts/srcinfo.sh write packages/pkgname/pkgname.pacscript
# generates .SRCINFO to packages/pkgname/.SRCINFO

# OR

list=($(ls packages/*/*.pacscript))
./scripts/srcinfo.sh write "${list[@]}"
# does it for the whole repo

./scripts/srcinfo.sh check
# checks if SRCINFOs exists for files and is tracked by git, for preventing merges without one
# can be run on individual files or array of files, assumes every file if none passed

./scripts/srcinfo.sh packagelist
# must be run from the head of the repository, updates the packagelist
```
- All `.SRCINFO` files now declare `pkgbase` at the top
- If a pacscript does not include `pkgbase`
    - `pkgbase` in `.SRCINFO` will default to the `pkgname`
    - `packagelist` entry is same as before
- If a pacscript *does* include `pkgbase`, then the `packagelist` entries are split up like:
```bash
${pkgbase}:pkgbase
${pkgbase}:${pkgname}
```
so for the example in https://github.com/pacstall/pacstall/pull/1153, our entries become:
```bash
rhino-pkg-git:pkgbase
rhino-pkg-git:rpk-git
rhino-pkg-git:rhino-pkg-git
```
This can of course also be done by hand, but either way is required for `pacstall` to handle `pkgbase` split packages.
They can then be called to install with any of the following:
```bash
pacstall -I rhino-pkg-git # same as with :pkgbase
pacstall -I rhino-pkg-git:pkgbase # prompts which to install
pacstall -I rhino-pkg-git:rpk-git # auto-selects only rpk-git split package
pacstall -I rhino-pkg-git:rhino-pkg-git # auto-selects only rhino-pkg-git split package
```
if using -Qa, this *must* be done without any `:extensions`:
```bash
pacstall -Qa "${pkgbase}#6143" # prompts which to install, will do all if none selected
pacstall -Qa rhino-pkg-git#1153 # as an example
```
Once installed on the system, you call split packages without their parent:
```bash
pacstall -Qi rpk-git
pacstall -R rpk-git
```


scripts/srcinfo.sh
```diff
@@ -20,28 +20,119 @@
...
+    pkgbuild_extract_to_srcinfo "$1" "${attr}" 0
+  done
+
+  for attr in "${multivalued[@]}"; do
+    pkgbuild_extract_to_srcinfo "$1" "${attr}" 1
+  done
+
+  get_pkgbuild_attribute "$1" 'arch' 1 'package_arch'
+  for a in "${package_arch[@]}"; do
+    [[ ${a} == any || ${a} == all ]] && continue
```
vigress8, line 113: So are we supporting `any` / `all` as an `arch` entry now? If so, is it not implicit anymore?

oklopfer: We already do support them, and they're still implicit. This is just saying if arch=any or all, don't try to append arch enhanced arrays

vigress8 (approved):

PR #6147, opened 2024-06-21
upd(tdrop-git): `1.5.0` -> `0.5.0`

erenfro:
Update tdrop-git to proper version, 0.5.0, and also added in missing mandatory dependencies (x11-utils), and optional dependencies, and fixed error on install routines not using ${pkgdir}.


Elsie19:
So since this is technically a downgrade, add epoch please.

oklopfer (approved):

thomascrha:
Thanks for fixing this @erenfro

erenfro:
> Thanks for fixing this @erenfro

Welcome! It's something I use so it's something in my high attention scope. :)

PR #6154, opened 2024-06-25
add: `typora-deb`

erenfro:
A minimal markdown editor and reader.

### DIFF
--- a/packages/typora-deb/typora-deb.pacscript
+++ b/packages/typora-deb/typora-deb.pacscript
@@ -1,18 +1,19 @@
 pkgname="typora-deb"
 gives="typora"
-pkgver="1.9.3"
+pkgver="1.9.3-1"
+url='https://typora.io'
 repology=("project: typora")
 depends=("libgtk-3-0" "libnss3")
 arch=('amd64')
 license=('custom:"Copyright (c) 2015 Abner Lee All Rights Reserved."')
 sha256sums=("deb47f0af1458d13e4cf6ee69b55ade61c204549cb2fb9692c5280da6a082f1f")
-source=("https://download.typora.io/linux/${gives}_${pkgver}_amd64.deb")
+source=("https://download.typora.io/linux/${gives}_${pkgver%-*}_amd64.deb")
 optdepends=(
-  "noto-fonts-enoji: Or some other emoji fonts to see/use emojis"
+  "noto-fonts-emoji: Or some other emoji fonts to see/use emojis"
   "pandoc: Import/export for extra file formats"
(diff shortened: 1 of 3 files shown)


erenfro:
Original typora deb file has:
Version: 1.9.3-1
Does this need to match up and reflect in the typora-deb pkgver, or is 1.9.3 fine for that?

Elsie19:
It's fine as is.

packages/typora-deb/typora-deb.pacscript
```diff
@@ -0,0 +1,18 @@
...
+arch=('amd64')
+license=('custom:"Copyright (c) 2015 Abner Lee All Rights Reserved."')
+sha256sums=("deb47f0af1458d13e4cf6ee69b55ade61c204549cb2fb9692c5280da6a082f1f")
+source=("https://download.typora.io/linux/${gives}_${pkgver}_amd64.deb")
+optdepends=(
+  "noto-fonts-enoji: Or some other emoji fonts to see/use emojis"
+  "pandoc: Import/export for extra file formats"
+)
+replaces=("typora")
+pkgdesc="A minimal markdown editor and reader."
```
Elsie19, line 15:
```suggestion
pkgdesc="A minimal markdown editor and reader"
```

packages/typora-deb/typora-deb.pacscript
```diff
@@ -0,0 +1,18 @@
+pkgname="typora-deb"
+gives="typora"
+pkgver="1.9.3"
+repology=("project: typora")
+depends=("libgtk-3-0" "libnss3")
+arch=('amd64')
+license=('custom:"Copyright (c) 2015 Abner Lee All Rights Reserved."')
```
Elsie19, line 7: This has to change. It's not a valid license name.

Elsie19 (changes requested):

Elsie19:
Match the deb version. `pkgver` should match exactly what is given in the deb's Version field.

packages/typora-deb/typora-deb.pacscript
erenfro: How so? This is proprietary payware software. It's not using any SPDX type licensing but their own.

oklopfer: then you don't need to provide the license variable at all (and shouldn't). It is meant for licenses that are already commonly available to systems through the `spdx-licenses` package (which, now that I say this out loud, should be made a pacstall dependency for next version).

oklopfer (approved):

PR #6168, opened 2024-06-29
fix(neovim): add in update-aternatives support

erenfro:
This adds in update-alternatives so neovim, view, and vimdiff can be registered and used in defaults.


Elsie19:
Can this be done for `neovim-git` as well?

packages/neovim/neovim.pacscript
```diff
@@ -1,11 +1,20 @@
 pkgname="neovim"
 pkgver="0.10.0"
-source=("https://github.com/neovim/neovim/archive/refs/tags/v${pkgver}.tar.gz")
+pkgrel="1"
```
Elsie19, line 3:
```suggestion
pkgrel="2"
```

Elsie19 (changes requested):

packages/neovim/neovim.pacscript
```diff
@@ -1,23 +1,82 @@
...
+  install -Dm755 "${srcdir}/ex" -t "${pkgdir}/usr/libexec/${pkgname}"
+  install -Dm755 "${srcdir}/rview" -t "${pkgdir}/usr/libexec/${pkgname}"
+  install -Dm755 "${srcdir}/rvim" -t "${pkgdir}/usr/libexec/${pkgname}"
+  install -Dm755 "${srcdir}/view" -t "${pkgdir}/usr/libexec/${pkgname}"
+  install -Dm755 "${srcdir}/vimdiff" -t "${pkgdir}/usr/libexec/${pkgname}"
+}
+
+post_install() {
+  set -e
+
```
Elsie19, line 47:
```suggestion
```

packages/neovim/neovim.pacscript
```diff
@@ -1,23 +1,82 @@
...
+  update-alternatives --install /usr/bin/rview rview "/usr/libexec/${pkgname}/rview" 30
+  update-alternatives --install /usr/bin/vi vi /usr/bin/nvim 30
+  update-alternatives --install /usr/bin/vim vim /usr/bin/nvim 30
+  update-alternatives --install /usr/bin/view view "/usr/libexec/${pkgname}/view" 30
+  update-alternatives --install /usr/bin/vimdiff vimdiff "/usr/libexec/${pkgname}/vimdiff" 30
+}
+
+pre_remove() {
+  set -e
+
```
Elsie19, line 62:
```suggestion
```

Elsie19 (changes requested):

packages/neovim/neovim.pacscript
erenfro: And why would you do this? This is actually intentionally set in this specific case.

packages/neovim/neovim.pacscript
erenfro: This too. This is intentional, and also based on the original neovim preinst control script.

PR #6195, opened 2024-07-04
add: `sleepy-launcher-bin`

vbrabandt2005:
- Added `sleepy-launcher-bin`
- Added the appropriate `.SRCINFO` file
- Added to `packagelist` (thank u pre-commit-ci)

I had forgot to modify the repology, but I had quickly fixed that, I had tested the pacscript locally as usual and it seems to be working great.


packages/sleepy-launcher-bin/sleepy-launcher-bin.pacscript
```diff
@@ -0,0 +1,44 @@
...
+incompatible=("ubuntu:bionic" "ubuntu:focal" "ubuntu:jammy" "ubuntu:kinetic" "debian:stretch" "debian:buster" "debian:bullseye")
+breaks=("${gives}" "${gives}-deb" "${gives}-app" "${gives}-git")
+
+optdepends=(
+  "mangohud: FPS Hud/GUI"
+  "gamemode: Game Optimizations"
+)
+
+package() {
+  cd "${_archive}"
```
Elsie19, line 25: Don't use _archive.

Elsie19 (changes requested):

packages/sleepy-launcher-bin/sleepy-launcher-bin.pacscript
vbrabandt2005: Sorry for being kinda lost, but I should just remove the `cd "${_archive}"` line entirely right? I was looking at the reference in the wiki and other people's commits, I see `${gives}` or `${pkgname}-${pkgver}` being used but for me I keep getting `/tmp/pacstall/bwrapenv.zOXikXkHJB: line 228: cd: sleepy-launcher (or -bin): Not a directory`

I saw PR #6140 which doesn't cd to anywhere and I assume mine should also be correct now

Elsie19: So, `${_archive}` was added as a measure to make the jump from pre-5.0.0 to 5.0.0 since we could add multiple items to download, instead of just having `url` which took one variable. It is deprecated for all new packages and we're in the process of removing it entirely.

vbrabandt2005: Oh that's good to know, but the script is ok now right? so I could set it for being ready to review

and in the future I will have to remove `${_archive}` for the other packages I maintain too correct?

Elsie19: Yeah you can set it for review, and yes, you should remove it from all packages you maintain when you have time. We will remove them for 6.0.0, but we'd prefer not to do *all* the work :)

vbrabandt2005: Nice, thanks for the info :)

oklopfer (approved):

PR #6196, opened 2024-07-04
upd(neovim-git): `0.9.1` -> `0.10.1`

erenfro:
Version bump, and added update-alternatives like `neovim`


packages/neovim-git/neovim-git.pacscript
```diff
@@ -1,21 +1,85 @@
...
+  install -Dm755 "${srcdir}/ex" -t "${pkgdir}/usr/libexec/${gives}"
+  install -Dm755 "${srcdir}/rview" -t "${pkgdir}/usr/libexec/${gives}"
+  install -Dm755 "${srcdir}/rvim" -t "${pkgdir}/usr/libexec/${gives}"
+  install -Dm755 "${srcdir}/view" -t "${pkgdir}/usr/libexec/${gives}"
+  install -Dm755 "${srcdir}/vimdiff" -t "${pkgdir}/usr/libexec/${gives}"
 }
+
+post_install() {
+  set -e
+
```
Elsie19, line 48:
```suggestion
```

packages/neovim-git/neovim-git.pacscript
```diff
@@ -1,21 +1,85 @@
...
+  update-alternatives --install /usr/bin/rview rview "/usr/libexec/${gives}/rview" 30
+  update-alternatives --install /usr/bin/vi vi /usr/bin/nvim 30
+  update-alternatives --install /usr/bin/vim vim /usr/bin/nvim 30
+  update-alternatives --install /usr/bin/view view "/usr/libexec/${gives}/view" 30
+  update-alternatives --install /usr/bin/vimdiff vimdiff "/usr/libexec/${gives}/vimdiff" 30
+}
+
+pre_remove() {
+  set -e
+
```
Elsie19, line 63:
```suggestion
```

Elsie19 (changes requested):

packages/neovim-git/neovim-git.pacscript
erenfro: And why would you do this? This is actually intentionally set in this specific case.

packages/neovim-git/neovim-git.pacscript
erenfro: This too. This is intentional, and also based on the original neovim preinst control script.

packages/neovim-git/neovim-git.pacscript
Elsie19: Because I can't think of a situation where this would ever fail unless something way more catestrophic situation happened.

erenfro: The intent is to literally report on ANY issues in the post-installation setup, because it's actually making system changes, this is critical. Without this, it would only report if the *last* command run were to exit with a non-zero return code. set -e, in this, insures it stops early, and reports post installation, and pre-removal issues that effect system changes. And it's also been in the Debian neovim package for 8 years and counting, for this reason.

erenfro: https://salsa.debian.org/vim-team/neovim/-/commit/db8850a951b1eb8085a2b182515b2f0884f105c1 -- For reverence.

packages/neovim-git/neovim-git.pacscript
Elsie19: I've just been informed that `set -e` is already in postscripts. You should remove these.

vigress8 (changes requested):
Considering these are such short scripts, I'd consider creating them inside the pacscript instead, following e.g. this pattern:
```bash
  install -Dm755 /dev/stdin "${pkgdir}/usr/libexec/${gives}/ex" <<'EOF'
#!/bin/sh
exec /usr/bin/nvim -e "$@"
EOF
```

packages/neovim-git/neovim-git.pacscript
erenfro: Okay. I'll get this removed.

erenfro:
> Considering these are such short scripts, I'd consider creating them inside the pacscript instead, following e.g. this pattern:
> 
> ```shell
>   install -Dm755 /dev/stdin "${pkgdir}/usr/libexec/${gives}/ex" <<'EOF'
> #!/bin/sh
> exec /usr/bin/nvim -e "$@"
> EOF
> ```

That's actually quite a good idea! Less sha256sums tracking as well! maybe a little less readability in the pacinstall script though as a result of it.

packages/neovim-git/neovim-git.pacscript
```diff
@@ -1,21 +1,81 @@
...
   make CMAKE_BUILD_TYPE=Release CMAKE_INSTALL_PREFIX=/usr -j"${NCPU}"
 }
 
 package() {
-  cd "${_archive}"
+  cd "${gives}"
   make install DESTDIR="${pkgdir}"
+  install -Dm644 "LICENSE.txt" -t "${pkgdir}/usr/share/licenses/${gives}"
+  install -Dm644 "README.md" -t "${pkgdir}/usr/share/doc/${gives}"
+  install -Dm755 /dev/stdin -t "${pkgdir}/usr/libexec/${gives}/ex" << 'EOF'
```
vigress8, line 24: `-t` is incorrect to use here, as it takes a directory as an argument. So this will create `/usr/libexec/${gives}/ex/stdin`

erenfro: D'oh! Good catch!

Elsie19 (approved):

PR #6197, opened 2024-07-04
upd(`btop-bin`): `1.2.13` -> `1.3.2`

pythoncrazy:
I also fixed it slightly, because it wasn't working correctly the first time I had tried it, due to not being in the correct directory. I also updated the version number, as two new versions had come out.


oklopfer:
this is a pacstall issue, we should be decompressing `.tbz`. I will make this fix for pacstall, but we will need to update this pacscript when a pacstall release comes out.

pythoncrazy:
Alright, I will hold off until then.

oklopfer (approved):

PR #6213, opened 2024-07-08
add: `libx86emu`

Xdavius:


packages/libx86emu/libx86emu.pacscript
```diff
@@ -0,0 +1,27 @@
+pkgname="libx86emu"
+gives=("libx86emu3")
+provides=("libx86emu-dev")
+pkgver="3.5"
+pkgdesc='x86 emulation library'
+arch=("amd64")
+url='https://github.com/wfeldt/libx86emu'
+pacdeps=("setconf")
+source=("git+https://github.com/wfeldt/libx86emu#commit=41aaa2f246d362f6d7dea262052a43141743ead2") # tag: 3.5
```
oklopfer, line 9: if this is a `tag`, then you should just use `#tag=${pkgver}` and not the commit hash.

oklopfer (changes requested):

packages/libx86emu/libx86emu.pacscript
Xdavius: It seems to be identical. I use your favorite one

oklopfer (approved):

PR #6230, opened 2024-07-11
fix(goverlay): updated to latest lazarus pacdeps

Xdavius:


oklopfer (changes requested):
You should bump `pkgrel` for this

Elsie19 (approved):

PR #6267, opened 2024-07-21
add: `lutris-git`

Xdavius:
pacdeps: "python3-moddb"


packages/lutris-git/lutris-git.pacscript
```diff
@@ -0,0 +1,75 @@
...
+  'winetricks: use system winetricks'
+  'libc6:i386: for 32bit games support'
+  'mangohud:i386: Display the games FPS + other information'
+  'libvkd3d1:i386: DirectX 12 support'
+  'mesa-vulkan-drivers:i386: lib32 Vulkan support')
+maintainer=("xdavius <xdavius@github.com>")
+
+prepare() {
+  cd "${gives}"
+}
```
oklopfer, line 62: not needed

packages/lutris-git/lutris-git.pacscript
```diff
@@ -0,0 +1,75 @@
...
+  'python3-pypresence: Discord Rich Presence integration'
+  'libvkd3d1: DirectX 12 support'
+  'vulkan-icd: Vulkan support'
+  'vulkan-tools: Vulkan support'
+  'wine: easiest way to get all the libraries missing from the Lutris runtime'
+  'winetricks: use system winetricks'
+  'libc6:i386: for 32bit games support'
+  'mangohud:i386: Display the games FPS + other information'
+  'libvkd3d1:i386: DirectX 12 support'
+  'mesa-vulkan-drivers:i386: lib32 Vulkan support')
```
oklopfer, line 57:
```bash
[+] INFO: Optional dependencies
[*] WARNING: mangohud:i386, libvkd3d1:i386, gamescope, python3-pypresence, linux-steam-integration does not exist in apt repositories
```

oklopfer (changes requested):

packages/lutris-git/lutris-git.pacscript
oklopfer:
```bash
[*] WARNING: gamescope, python3-pypresence, winetricks, linux-steam-integration does not exist in apt repositories
```
hmm im going to look into these

oklopfer:
```bash
[*] WARNING: winetricks, python3-pypresence, linux-steam-integration does not exist in apt repositories
```
for sid/trixie

Xdavius: > for unstable + testing (looks like our CI does not use contrib)

Contrib is not needed for Debian. No deps error @home in Stable or Testing ....

Xdavius: I dot it. For ubuntu : optdeps i386 -> i386 is disabled for ubuntu. need to activate arch=("amd64" "i386") ??

Xdavius: > python3-pypresence + linux-steam-integration
> 
> these are the issue ones, because they don't exist in any repository
> 
> the other ones are fine because they exist in some and that's why the optdeps warning allows continue and not error

They are not failling. optdepends doesn"t search for them if they doesn't exists isn't it ? (my new commit doesn't give errors, now lutris run well !)

I will work on theses : 

gamescope, python3-pypresence, linux-steam-integration

I don't want to discard some optdepends. I want the full capacity of the package =)

packages/lutris-git/lutris-git.pacscript
```diff
@@ -0,0 +1,73 @@
+pkgname="lutris-git"
+gives="lutris"
+pkgver="0.5.17"
+pkgdesc="Open Gaming Platform"
+url='https://lutris.net'
+license=("GPL-3.0-or-later")
+provides=("${gives}-git")
+replace=("${gives}-deb" "${gives}")
```
oklopfer, line 8: replaces*

oklopfer (changes requested):

packages/lutris-git/lutris-git.pacscript
oklopfer: Well we should discard the optdeps that do not exist in apt repos, since they are useless. Maybe they can be pacdeps, but they shouldn't be somewhere where they don't exist

Xdavius: It's becoming a feature request: optpacdeps 😜

oklopfer: I remember also we can use `optdepends_ubuntu` and `optdepends_debian`

Xdavius: > I remember also we can use `optdepends_ubuntu` and `optdepends_debian`

Tomorrow I will add gamescope in optdepends_debian

The 2 others I need optpacdeps but this option need to be seted as No by default whereas optdeps are set as Yes.

The good thing is that if the optdepend doesn't exists, it's ignored by default and not available to users.

I think we can go ahead as this and fix when the option will be available.

No doubts also that most people will not install the git version but the -deb one.

oklopfer:
Was there something wrong with optdepends_debian?

Xdavius:
> Was there something wrong with optdepends_debian?
> 
> EDIT: appears yes, will look into today hope

Discord me when it's good plz. So I add for debian. Of I can add a not sexy IF ^^

Xdavius:
@oklopfer still missing something for this package ? For me it's seems all good. Can you check?

oklopfer (approved):

PR #6273, opened 2024-07-22
add: `jsoncpp`

Xdavius:


packages/jsoncpp/jsoncpp.pacscript
```diff
@@ -0,0 +1,53 @@
+pkgname="jsoncpp"
+gives="libjsoncpp25"
+provides=('libjsoncpp-dev' 'libjsoncpp.so')
+conflicts=('libjsoncpp25' 'libjsoncpp-dev' 'libjsoncpp.so')
```
oklopfer, line 4: .so is a library file name, not a package. We don’t use those in pacstall (at least, not yet)

packages/jsoncpp/jsoncpp.pacscript
```diff
@@ -0,0 +1,53 @@
...
+
+prepare() {
+  cd "${pkgname}-${pkgver}"
+  echo "${pkgver}" > version
+}
+
+build() {
+  cd "${pkgname}-${pkgver}"
+  echo "Building shared library..."
+  meson --prefix=/usr/local --default-library shared build
```
oklopfer, line 23: Why not regular /usr?

oklopfer (changes requested):

packages/jsoncpp/jsoncpp.pacscript
Xdavius: > Why not regular /usr?

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1076787

Because in Debian/Ubuntu some packages are packaged in non standard way to avoid problems. 
We should install pacstall libs in /usr/local to avoid conflicts with systemlibs. Seems to be a better practice.

I will set all my future lib packages in /usr/local. Normaly if no prefix is defined, it should be the default path, /usr prefix is reserved to system libs, not foreign libs.

packages/jsoncpp/jsoncpp.pacscript
```diff
@@ -0,0 +1,54 @@
+pkgname="jsoncpp"
+gives="libjsoncpp-dev"
+provides=("libjsoncpp25")
+replaces=("libjsoncpp25" "libjsoncpp-dev")
+breaks="${replaces[*]}"
```
oklopfer, line 5: This needs ()

oklopfer (changes requested):

Xdavius:
Closing. No more needed to test.

PR #6277, opened 2024-07-24
fix(xcb-util-errors): replaces/conflicts/provides

Xdavius:


Xdavius:
@Elsie19 what do you think like this ?
It should be fine no ?

oklopfer (approved):

PR #6288, opened 2024-07-26
add: `apptainer` and `apptainer-suid`

D-Brox:
Required for https://github.com/rhino-linux/rhino-setup/pull/68


Elsie19:
One package per PR.

PR #6339, opened 2024-08-16
add: `nerd-fonts`

oklopfer:
because of how `pkgbase` splitting is currently written, each `pkgname` is downloading all of the sources, resulting in 65x831MB downloads. We have to fix this. This will probably be blocked for a long time.


Elsie19:
Is this ready to test now?

oklopfer:
> Is this ready to test now?

I gotta fix the package functions here as they don’t grab from the right place right now (all using */)

PR #6355, opened 2024-08-22
add: `libcs50-git`

vvaibhavv11:
Add new package libcs50-git

This is my first time to do any contribution to any project.

### DIFF
--- a/packages/libcs50-git/libcs50-git.pacscript
+++ b/packages/libcs50-git/libcs50-git.pacscript
@@ -14,10 +14,6 @@ build() {
 package() {
   cd "${_archive}"
   make DESTDIR="${pkgdir}/usr/local" install || true
-
-  # Manually skip or suppress the ldconfig error
-  # Skipping ldconfig to avoid read-only filesystem issues
-  echo "skipping the ldconfig because it gives error doing this after the install"
 }
 
 post_install() {
--- a/packages/libcs50-git/libcs50-git.pacscript
+++ b/packages/libcs50-git/libcs50-git.pacscript
@@ -8,7 +8,7 @@ pkgdesc="Library provided by Harvard University for learning the basic C languag
 
 build() {
   cd "${_archive}"
-  make
+  make -j"${NCPU}"
 }
 
 package() {
--- a/packages/libcs50-git/libcs50-git.pacscript
+++ b/packages/libcs50-git/libcs50-git.pacscript
@@ -7,12 +7,12 @@ maintainer=("Vaibhav Sharma <vvaibhavv3434@gmail.com>")
 pkgdesc="Library provided by Harvard University for learning the basic C language"
 
 build() {
-  cd "${_archive}"
+  cd "${gives}"
   make -j"${NCPU}"
 }
 
 package() {
-  cd "${_archive}"
+  cd "${gives}"
   make DESTDIR="${pkgdir}/usr/local" install || true
 }
 
--- a/packages/libcs50-git/libcs50-git.pacscript
+++ b/packages/libcs50-git/libcs50-git.pacscript
@@ -1,0 +1,25 @@
+pkgname="libcs50-git"
+gives="libcs50"
+pkgver="11.0.3"
+source=("https://github.com/cs50/libcs50.git")
+makedepends=("make")
+maintainer=("Vaibhav Sharma <vvaibhavv3434@gmail.com>")
+pkgdesc="Library provided by Harvard University for learning the basic C language"
+
+build() {
+  cd "${_archive}"
+  make
+}
+
+package() {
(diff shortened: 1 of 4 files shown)


packages/libcs50-git/libcs50-git.pacscript
```diff
@@ -0,0 +1,25 @@
...
+gives="libcs50"
+pkgver="11.0.3"
+source=("https://github.com/cs50/libcs50.git")
+makedepends=("make")
+maintainer=("Vaibhav Sharma <vvaibhavv3434@gmail.com>")
+pkgdesc="Library provided by Harvard University for learning the basic C language"
+
+build() {
+  cd "${_archive}"
+  make
```
Elsie19, line 11:
```suggestion
  make -j"${NCPU}"
```
We add this for parallel building.

packages/libcs50-git/libcs50-git.pacscript
```diff
@@ -0,0 +1,25 @@
+pkgname="libcs50-git"
+gives="libcs50"
+pkgver="11.0.3"
+source=("https://github.com/cs50/libcs50.git")
+makedepends=("make")
+maintainer=("Vaibhav Sharma <vvaibhavv3434@gmail.com>")
+pkgdesc="Library provided by Harvard University for learning the basic C language"
+
+build() {
+  cd "${_archive}"
```
Elsie19, line 10: Please remove this line and anywhere else and replace it with something else, as this variable id deprecated.

packages/libcs50-git/libcs50-git.pacscript
```diff
@@ -0,0 +1,25 @@
...
+  make
+}
+
+package() {
+  cd "${_archive}"
+  make DESTDIR="${pkgdir}/usr/local" install || true
+
+  # Manually skip or suppress the ldconfig error
+  # Skipping ldconfig to avoid read-only filesystem issues
+  echo "skipping the ldconfig because it gives error doing this after the install"
```
Elsie19, line 20:
```suggestion
```

Elsie19 (changes requested):

packages/libcs50-git/libcs50-git.pacscript
vvaibhavv11: done changes

packages/libcs50-git/libcs50-git.pacscript
```diff
@@ -0,0 +1,21 @@
...
+pkgdesc="Library provided by Harvard University for learning the basic C language"
+
+build() {
+  cd "${gives}"
+  make -j"${NCPU}"
+}
+
+package() {
+  cd "${gives}"
+  make DESTDIR="${pkgdir}/usr/local" install || true
```
oklopfer, line 16: why is `|| true` being used here?

oklopfer (changes requested):

packages/libcs50-git/libcs50-git.pacscript
vvaibhavv11: Because of the ldconfig it presents in the make file when it's fail to run this command `ldconfig $(DESTDIR)/lib` it just gives error and the pacstall says package is not installed, so I added the `|| true` so it does not give error and I add the ldconfig command in the post_install function.

vvaibhavv11: And the rest of the script is just the cleanup like delete the build folder and stuff, so I can just ignore the error and move forward.

vvaibhavv11: Next time if I do a thing like that I just add a comment and that fine right, and you can just delete the comment later if you want to.

vvaibhavv11: > Because of the ldconfig it presents in the make file when it's fail to run this command `ldconfig $(DESTDIR)/lib` it just gives error and the pacstall says package is not installed, so I added the `|| true` so it does not give error and I add the ldconfig command in the post_install function.

i think i give the read only file system error ` [!] ERROR: Could not package libcs50-git properly
`.

vvaibhavv11: > > Because of the ldconfig it presents in the make file when it's fail to run this command `ldconfig $(DESTDIR)/lib` it just gives error and the pacstall says package is not installed, so I added the `|| true` so it does not give error and I add the ldconfig command in the post_install function.
> 
> i think i give the read only file system error `[!] ERROR: Could not package libcs50-git properly`.

and this also `/sbin/ldconfig.real: Can't create temporary cache file /etc/ld.so.cache~: Read-only file system
` by the make file.

vigress8 (approved):

Elsie19 (approved):

PR #6371, opened 2024-08-30
upd(linux-kernel-stable): `6.10.4` -> `6.10.6`

oklopfer:


Xdavius:
I think all kernel pacscripts have to be incompatible debian:*

I will avoid mistakes for users.

PR #6387, opened 2024-09-04
add: `osu-lazer-app`

AozoraDev:
Click that circle!

EDIT: osu-mime coming soon.


AozoraDev:
Yep it takes too long. Gotta maintain it by myself then.

PR #6404, opened 2024-09-10
add: `lib32-wayland`

Xdavius:
DO NOT MERGE : need chroot i386 clean build


Xdavius:
checks are in amd64, but package is only i386.

Xdavius:
not supported

PR #6420, opened 2024-09-12
fix(apx-git): declare `incompatible` with bookworm

souhaiebtar:
to declare incompatible with bookworm


Xdavius:
> to declare incompatible with bookworm
 Change PR title 

fix(apx-git): declare incompatible with bookworm

vigress8 (approved):

PR #6422, opened 2024-09-13
add: `ffnvcodec-headers`

luigi311:
First contribution so let me know if something looks wrong


vigress8 (approved):
I edited the title to match our expected format, for the future.
In any case, thanks for the contribution!

PR #6438, opened 2024-09-21
upd(librewolf-app): `125.0.3-1` -> `130.0-3`

jimed-rand:
I re-commit what I did earlier but now on branch `pr-branch-2` for updating `librewolf-app` from `125.0.3-1` --> `130.0-3`. Please review it immediately when everything's okay. Also, I want evaluate it before merged when got something wrong.


packages/librewolf-app/librewolf-app.pacscript
```diff
@@ -48,7 +51,7 @@ package() {
 post_remove() {
   # Delete ~/.librewolf if it exists
   if [[ -d "${homedir}/.librewolf" ]]; then
-    if ask "Do you want to delete ~/.librewolf/ files?" Y; then
+    if ask "Did you want to delete ~/.librewolf/ files?" Y; then
```
vigress8, line 54:
```suggestion
    if ask "Do you want to delete ~/.librewolf/ files?" Y; then
```
`Do` makes more sense here

jimed-rand: Okay, I'll revise it immediately

vigress8 (approved):

Issue #1726, opened 2022-11-04
PacReq: Ungoogled Chromium


brainchild0:
### Types

-deb (Debian achieve release of the package)

### URL

https://github.com/ungoogled-software/ungoogled-chromium

### Building guide

https://github.com/ungoogled-software/ungoogled-chromium/blob/master/docs/building.md

### Additional context

Debian package from custom repository had worked on versions of LTS Ubuntu that are now outdated, but the repository no longer is maintained with a working key or populated with current builds.

jimed-rand:
I am sorry to reply it this time on 2024 while you requested 2 years ago.

I could add it the `deb` version, but only this version (`100.0.4896.127-1`) could be added for that. Is that okay to add it this version ?

jimed-rand:
https://github.com/berkley4/ungoogled-chromium-debian/releases

Now I got the newest sources for `ungoogled-chromium-deb`, I'll use this as new base.

jimed-rand:
Now commit #6453 has been live on official repo @brainchild0 
You can close this issue as complete.

PR #6440, opened 2024-09-21
add: `ungoogled-chromium-deb`

jimed-rand:
This commit is based on request #1726.


jimed-rand:
Before merge this commit, I suggest to merge first commit #6441 cause it's important dependences for `ungoogled-chromium-deb` or #6440

jimed-rand:
Now it's back to review mode and now #6443 are replacing the #6441

vigress8:
This Chromium version is years old and unsupported. Please use the latest available version

jimed-rand:
> This Chromium version is years old and unsupported. Please use the latest available version

I know it's kinda old, but that's only one `deb` version available. But for `app` version, it has newest version.

jimed-rand:
I should convert back into draft and find the newest version first.

jimed-rand:
I got the newest version @vigress8 
Ready to review again.

jimed-rand:
After tests, the package are not compatible and need more depends on `debian:stable` which it still uses Bookworm, but there's okay on another version. I'll add `debian:stable' as incompatible flag

PR #6441, opened 2024-09-21
add: `ungoogled-chromium-common-deb`

jimed-rand:
Same as what I'd commit on #6440 which based on request #1726. But this commit are the dependences for `ungoogled-chromium-deb` which still reviewing on #6440.

PR #6443, opened 2024-09-21
add: `ungoogled-chromium-common-deb`

jimed-rand:
Now I had re-commit `ungoogled-chromium-common-deb` on `pr-branch-4`. Same as what had been canceled on #6441 but on new branch.


jimed-rand:
I found the new sources, but `common` has been obsolete for newest version, so I will close this as "obsolete" package and `pr-branch-4` will be deactivated branch.

PR #6448, opened 2024-09-22
upd(microsoft-edge-dev-deb): `118.0.2060.1` -> `130.0.2835.2-1`

jimed-rand:
Just another update of Microsoft Edge Dev from `118.0.2060.1` --> `130.0.2835.2-1`. Please merge immediately when everything's okay.


vigress8:
Don't update .SRCINFO manually, that can mess things up.
If you install [`pre-commit` ](https://pre-commit.com/#install) and run `pre-commit install`, the .SRCINFO will always be autogenerated for you

jimed-rand:
> Don't update .SRCINFO manually, that can mess things up. If you install [`pre-commit`](https://pre-commit.com/#install) and run `pre-commit install`, the .SRCINFO will always be autogenerated for you

Ah, I see. that's why precommit doesn't work. Anyway, I commit it manually and not using `git` on Terminal, and I was think that `.SRCINFO` was generaten manually. So, I just need to write only pacscript, right ?

PR #6453, opened 2024-09-23
add: `ungoogled-chromium-deb`

jimed-rand:
In previous commit due to deleted previous fork, I re-commit again with sources from @berkley4 and now had been good at pre-commit before commit into my own forks. Please review and merge immediately. The version I'd commit is stable version, while previous commit was unstable version. Be careful.


jimed-rand:
I forget, it also based on request #1726 which now on progress.

PR #6498, opened 2024-10-03
upd(brave-{browser,keyring}-deb): `1.66.118` -> `1.70.123`, `1.13` -> `1.17`

bibelin:
- `brave-browser-deb` wasn't updated in 4 months. I constantly use Brave browser, so I'm interested in keeping this package up to date and I added myself as maintainer, I hope it's okay.
- Added `arm64` support
- I don't know if there's a reason to use github links for deb sources over everything else, but I can only speak for the opposite: for some people (myself included, unfortunately) downloading from github is usually painfully slow, even tho generally network speed is okay. For example [here](https://reddit.com/r/github/comments/ekvvff/extremely_slow_downloads_from_github/) are complains about it from 5 to 2 years ago, and I know that these are from a long time ago and that's just a bunch of people, but the fact that I still experience the same problem makes me think that there still may be other pacstall users wanting to download Brave but having the same issue. And changing the source fixes it (in my case, ~30mins from github vs <1min from brave deb repo). So I hope you'll find this change justified. After all, that's also official Brave source and it was already used in `brave-keyring-deb`.


Elsie19:
There is a reason why we use github links and that is because it allows people to more easily see where the source is coming from. And either way, if we use their upstream repo to download, we might as well just use their apt repo instead of pacstall.

I also don't know why GitHub would be slow for you. Is this a consistent issue or can you confirm it's the same on other devices/networks?

bibelin:
> I also don't know why GitHub would be slow for you. Is this a consistent issue or can you confirm it's the same on other devices/networks?

Browsing Github website in web browser isn't slow, but downloading files (in browser or with CLI tools, doesn't matter) and `git clone` from github (but not from other services) are very slow, <100Kb/s usually. I found some settings to help with cloning, but it's not ideal, cloning at least relatively big repo usually fails. And it's always like that, on different devices and with different networks.
It's either ISPs (plural - I tried different ones) throttling some parts of Github, or Github itself adding some weird restrictions, which doesn't seems like a surprise to me, I had another issue with unexpected Github behavior and lies about its reason, but that's unrelated.

> that is because it allows people to more easily see where the source is coming from

Well sure `https://brave-browser-apt-release.s3.brave.com` looks more cryptic but the ones who care about will see `brave.com`, no?

bibelin:
> And either way, if we use their upstream repo to download, we might as well just use their apt repo instead of pacstall.

I would say that from that point of view objectively it doesn't make sense to create a pacscript for any program that can be installed from universal deb repo. But there are some, because I guess different users can have not so obvious reasons. Like in my case I don't want to make KDE Discover, which I use to update system with all deb packages, even more slower by adding more repos, and delegate to pacstall updating some apps.

Elsie19:
> I understand you can disagree, but I see a reason to maintain the package even if it's just a replacement for the deb repo.

I know you haven't been a part of pacstall for a long time, but that's an unofficial official rule. We don't use deb repos as sources unless it serves a highly specific purpose (think pinned versions).

> Browsing Github website in web browser isn't slow, but downloading files (in browser or with CLI tools, doesn't matter) and git clone from github (but not from other services) are very slow, <100Kb/s usually.

That doesn't have anything to do with brave deb. You must have a network issue because you're the first person I've heard say that.

bibelin:
> That doesn't have anything to do with brave deb. You must have a network issue because you're the first person I've heard say that.

Well... You asked about my issue with github, I answered and said that there are - or at least were - people with the same problem. And then I said:

> And please don't tell me about proxy or VPN, I know that it could fix the issue for me, but we are talking about pacstall users from all the world wanting to install a package, not about my issues with Github.

My point is that if not for existing then for possible future users of Pacstall it will be better to actually avoid Github file downloads when it's possible to use other **official** sources. It doesn't matter whether my issue is caused by github or my government. Imagine a pacstall user who will try to install a package that downloads files from Github instead of another official source. And because of either Github, the user's ISP or the user's government the download speed will be slow. And in this particular case, in case of brave package, maybe changing the source will fix the problem for that particular user. And it's not just about installing, but also about updating.
Sure, your repo - your rules, but I don't understand why to not break the rule if it leads to a benefit **to your users**, and without downsides.

> I know you've sent something on discord about having network issues with external_connection=true, so maybe that is part of your problem.

That's completely unrelated.

oklopfer:
This seems to have gotten off track, and you both seem to have gone over each other's points. 
1. At the end of the day, the GitHub sources are still primary release sources, and the apt links are secondary mirrors. They are equally as "official." 
It should really just come down to which is more reliable:
    - in download speeds
    - in url structure/organization
    - and in consistency of maintaining said organization
2. That being said, I can attest to users having issues with downloading from GitHub vs other sources, so it does seem reasonable to make this change.
Readability between:
https://brave-browser-apt-release.s3.brave.com/pool/main/b/brave-browser/brave-browser_1.70.123_amd64.deb
and
https://github.com/brave/brave-browser/releases/download/v1.70.123/brave-browser_1.70.123_amd64.deb
is barely a change, they're both pretty obvious.

vigress8 (approved):

PR #6504, opened 2024-10-06
upd(librewolf-deb): `130.0-3` -> `130.0.1-1`

jimed-rand:
This time for fixing sources of `librewolf-deb` due to migration issues by LibreWolf developer and merge Debian builds into `http://repo.librewolf.net` [This article might explain what is that mean and what I'd do in this commit request.](https://librewolf.net/debian-migration/)


jimed-rand:
Also, I've downgraded for versions from `130.0.3` into `130.0.1-1`

If downgrade are not allowed, I'll fix the sources again from official repo into their GitLab repos

jimed-rand:
Oops, I was make a mistaken. Alright, I'll close this as mistaken and re-open as `upd`, not an `fix`

PR #6519, opened 2024-10-10
upd(intellij-idea-community-bin): `2024.1` -> `2024.2.3`

jimed-rand:
Same as #6518, but it's IDEA.


jimed-rand:
SHA256 for `arm64` are not ready yet. So, until I add commit for fixing the SHA256 sum for `arm64`, this commit will remain draft until it finished.

Elsie19 (approved):

PR #6520, opened 2024-10-10
upd(vscodium-deb): `1.92.1.24225` -> `1.94.0.24282`

jimed-rand:
Update the VSCodium from previous version into latest. Merge immediately when everything's okay.


jimed-rand:
I'll recommit due little mistakes

vigress8:
What was the mistake?

jimed-rand:
> What was the mistake?

I forgot something on my mind and I don't see labels, that's why I'd re-commit

PR #6534, opened 2024-10-14
add: `orchis-kde`, `qt6-style-kvantum`

bibelin:
## Progress 

- [x] Edit packagelist
- [x] Add initial pacscript
- [ ] Contact devs

## Notes about `orchis-kde`

This package [exists](https://packages.ubuntu.com/oracular/orchis-kde) in Ubuntu, but only for noble and oracular and is outdated. A theme repo doesn't have tags or releases, although new commits appear very rarely, so each new commit can be considered a new release. Ubuntu package uses `YYYY-MM-DD` version format, so did I.

## Notes about `qt6-style-kvantum`

This package [exists](https://packages.ubuntu.com/oracular/qt6-style-kvantum) in Ubuntu oracular (thus I used this name instead of updating unmaintained `kvantum` package, which I suppose could be removed if such operation could be done in this repo: if there still are packages that depend on it, `qt5-style-kvantum | qt6-style-kvantum` could be used instead). But required dependencies exist in Debian testing and sid, and in KDE Neon. In Ubuntu repo the package is split in `${pkgname}`, `${pkgname}-themes` and `${pkgname}-l10n`, but I kept everything in one package for user convenience (users probably want l10n anyway, it's just a few Kb, and themes take a few Mb, so I doubt it can be considered a bloat and so I avoided extra work,  but if you insist I will make a split package).

----

UPD: I see tests failing because of `qt5-style-kvantum` is installed. I checked that both qt5 and qt6 versions exist in Ubuntu oracular, yet both have `/usr/bin/{kvantummanager,kvantumpreview}` so they should be in conflict. :thinking: Any suggestions?


bibelin:
I realized the problem with tests was that `qt5-style-kvantum` gets installed with `orchis-kde` because that is its optional dependency, and this blocks installation of `qt6-style-kvantum`. I will open two separate PRs. Sorry for inconvenience! :pray:

PR #6604, opened 2024-10-30
fix(`waydroid-deb`): Fix the invalid SHA256 sums and change the descriptions

jimed-rand:
Nothing to tell it, just merge it immediately


jimed-rand:
Anyway, this commit are not updating the versions, but it replaces the invalid SHA256 (because the file has been updated in the sources), and also add severe words in `pkgdesc`. Updating on this commit means fixing the problems I'd face it. Please merge immediately

jimed-rand:
If in December are still not been merged, I'll close it as unplanned.

oklopfer (approved):

PR #6610, opened 2024-11-01
upd(zen-browser): 1.0.1-a.16 -> 1.0.1-a.17

D-Brox:


Elsie19:
This should've had quotes around the versions.

Issue #6586, opened 2024-10-26
PacReq: Linux kernel LTS 6.6


jimed-rand:
### Types

-deb (Debian packages)

### URL

https://kernel.ubuntu.com/mainline/v6.6.51/ (Sources)

### Building guide

I'll create 3 commits (2 dependencies and main package) and here's merge map below

`linux-longterm-6.6-modules-deb` --> `linux-longterm-6.6-headers-deb` --> `linux-longterm-6.6-deb`

After the end of merge map has been reached and all has been merged, Close the request as all commit had been merged.

### Additional context

The latest LTS kernel today are `6.6.58`, but only `6.6.51` avaliable on sources.

jimed-rand:
To save time, for the `packagelist` file I will write it all at once as below.

``sh
linux-longterm-6.6-deb
linux-longterm-6.6-headers-deb
linux-longterm-6.6-modules-deb
```

I commit this section in the commit for `linux-longterm-6.6-modules-deb`, again to save time.

jimed-rand:
So, for commit `linux-longterm-6.6-headers-deb` and `linux-longterm-6.6-deb`, merge it immediately after merging `linux-longterm-6.6-modules-deb`.

jimed-rand:
Something's wrong, I'll convert #6611 and #6612 into draft. So don't touch the commit unless it's ready to commit

jimed-rand:
Commit #6611 ready to review. And also, #6612 will keep being drafted until #6611 has been merged.

jimed-rand:
Alright, new plan. After `linux-longterm-6.6-modules-deb` has been merged, I revise the merge map for the `headers` section. Here's the merge map below.

`linux-longterm-6.6-headers-modular-deb` -> `linux-longterm-6.6-headers-deb` -> `linux-longterm-6.6-deb`

For this (`linux-longterm-6.6-headers-modular-deb`), it's actually kernel headers with `all` flags because it supports both `amd64` and `arm64`, so for the `arch` flag are `all` instead `any`. So, follow this merge map and after `linux-longterm-6.6-deb` has been merged, you can close this requests issues as completed.

jimed-rand:
#6612 closed as unplanned because I got missing dependencies. After looking what's missing, I guess it could need new merge map.

oklopfer:
Completed in https://github.com/pacstall/pacstall-programs/commit/d7d1724430723e8436cf0b2e06b6ab0b46267545

PR #6611, opened 2024-11-02
add: `linux-longterm-6.6-modules-deb`

jimed-rand:
Based in #6586 requests. Also, see the requests firsts and re-check the commit I did before merging on it

PR #6612, opened 2024-11-02
add: `linux-longterm-6.6-headers-deb`

jimed-rand:
Based on #6586 requests and this commit are 2nd tree of #6611. Merge this commit immediately when #6611 are okay to merge.


jimed-rand:
I guess it need another dependence, alright. I'll close it as unplanned and re-plan the merge tree for this, but please don't close the issues #6586

PR #6619, opened 2024-11-08
add: `linux-headers-lts-6.6-deb`

jimed-rand:
Continuation of these requests (#6586)


jimed-rand:
It still got errors when I made a change. Should I close it as unplanned and re-commit with new branch or keep it be?

vigress8 (dismissed):

packagelist
```diff
@@ -299,6 +299,7 @@ linux-image-unsigned-stable-deb
 linux-kernel
 linux-kernel-stable
 linux-libc-dev-okpine-deb
+linux-longterm-6.6-headers-modular-deb
```
oklopfer, line 302: drop `-modular` from the naming of the package

oklopfer (changes requested):

packagelist
oklopfer: Actually the other one needs to be changed too. 
They should be:
`linux-headers-lts-6.6-deb` and `linux-modules-lts-6.6-deb`

jimed-rand:
> What’s with the `-modular` naming? It is confusing and does not match the standard of other kernels we package

`modular` means it's primary headers for all architecture, it's located on `amd64` and if you see the sources, there's one primary headers and second headers which it's `generic`.

oklopfer:
> > What’s with the `-modular` naming? It is confusing and does not match the standard of other kernels we package
> 
> `modular` means it's primary headers for all architecture, it's located on `amd64` and if you see the sources, there's one primary headers and second headers which it's `generic`.

See https://github.com/pacstall/pacstall-programs/pull/6619#discussion_r184097586, we have a standard naming practice for these

packagelist
jimed-rand: > Actually the other one needs to be changed too. They should be: `linux-headers-lts-6.6-deb` and `linux-modules-lts-6.6-deb`
> 
> FWD from discord: So in the end based on how Ubuntu packages mainline kernel, your pacscripts should be:
> 
> ```
> linux-headers-lts-6.6-deb
> linux-headers-generic-lts-6.6-deb
> linux-image-unsigned-lts-6.6-deb
> linux-modules-lts-6.6-deb
> ```

So, the merge map should be:

`linux-modules-lts-6.6-deb` -> `linux-headers-lts-6.6-deb` -> `linux-headers-generic-lts-6.6-deb` -> `linux-image-unsigned-lts-6.6-deb`

Like that, right ?

jimed-rand:
> > > What’s with the `-modular` naming? It is confusing and does not match the standard of other kernels we package
> > 
> > 
> > `modular` means it's primary headers for all architecture, it's located on `amd64` and if you see the sources, there's one primary headers and second headers which it's `generic`.
> 
> See [#6619 (comment)](https://github.com/pacstall/pacstall-programs/pull/6619#discussion_r1840975863), we have a standard naming practice for these

Alright, I follow what you'd mean. So, should I fix the `modules` and this merged commit in next commit yeah ?

packagelist
oklopfer: There’s no reason why they can’t all be uploaded at once. See how `linux-kernel` and `linux-kernel-stable` are packaged for reference, you could create `linux-kernel-lts-6.6` meta package and do them all in one go

jimed-rand: > There’s no reason why they can’t all be uploaded at once. See how `linux-kernel` and `linux-kernel-stable` are packaged for reference, you could create `linux-kernel-lts-6.6` meta package and do them all in one go

So, what should I do know for this commit and the last commit before for `modules`? _I am still confused because I replied this in the waking hours_

jimed-rand: Continue on Discord for replying on it

jimed-rand: Got it, it's fixed. Please accept it when `pre-commit` get the job done.

PR #6622, opened 2024-11-09
Upd: cortile version

ajstrongdev:
Update the cortile version and remove automatic updates by utilising `cortile disable-release-info`

So far i've incremented the pkgver, however I do not understand how to disable automatic updates through a pacscript, making this a draft pull request now until completed.


ajstrongdev:
Misunderstood what was required, i need to create cortile-git instead.

PR #6710, opened 2024-11-30
add: `rpi-imager`

sundaram123krishnan:


jimed-rand:
I guess the building on the logs are failed anyway

jimed-rand:
Anyway @oklopfer and @Elsie19, close this commit as unplanned if still Stalled. I want to make the commit for this `rpi-imager`.

sundaram123krishnan:
locally it works fine, but i unable to pass the CI @jimed-rand 
there is some issue with the qt installation

jimed-rand:
> locally it works fine, but i am unable to pass the CI @jimed-rand there is some issue with the qt installation

Well, you make it using Git and no PacReq refrences (before I've finally add it), you should make it easier with changes the pacscript from `git` build into `app` or AppImage instead. It's making save a times.

PR #6741, opened 2024-12-09
upd(legcord-deb): `1.0.3` -> `1.0.5`

jimed-rand:
Just usual updates for Legcord


jimed-rand:
Due to failures, I'll close it as unplanned and re-commit again later

PR #6720, opened 2024-12-05
fix(unicorn-{desktop,mobile}-git): add dynamic workspaces 

ajstrongdev:
This pacscript update will bump dynamic workspaces from Stampede (testing) into the Unicorn pacscript as a pacdep

PR #6750, opened 2024-12-11
fix(xfce4-dynamic-workspaces-plugin-git): change `ln` to `cp`

ajstrongdev:
https://github.com/pacstall/pacstall-programs/pull/6720#issuecomment-2525302732


Elsie19 (approved):

oklopfer:
Nice job reviewing. `cp -sf` means it is still a symlink.

PR #6759, opened 2024-12-13
add: `zen-browser-aarch64-bin`

D-Brox:


D-Brox:
The name is aarch64 because that's their naming convention
image

Issue #6357, opened 2024-08-22
Broken: name_of_the_package


iconoclasthero:
### Package name

wine-staging

### Version

9.8??  whatever rpk pulled as the current one

### Pacstall version

5.4.1 Telemagenta

### Relevant log output

```shell
$ rpk install wine
Found packages matching 'wine':

[0]: carla-bridge-wine64 (apt)
[1]: fonts-wine (apt)
[2]: gnome-wine-icon-theme (apt)
[3]: libkwineffects14 (apt)
[4]: librust-winapi-dev (apt)
[5]: libwine (apt)
[6]: libwine-dev (apt)
[7]: q4wine (apt)
[8]: shiki-wine-theme (apt)
[9]: twine (apt)
[10]: wine (apt)
[11]: wine-binfmt (apt)
[12]: wine-stable (apt)
[13]: wine64 (apt)
[14]: wine64-preloader (apt)
[15]: wine64-tools (apt)
[16]: winetricks (apt)
[17]: wine32 (apt)
[18]: wine32-preloader (apt)
[19]: wine32-tools (apt)
[20]: carla-bridge-wine32 (apt)
[21]: winetricks-git (pacstall)
[22]: wine-staging-git (pacstall)

Select which package to install [0-22]: 22
Selecting 'wine-staging-git' from package manager 'pacstall'
Are you sure? (Y/n) 
(wine-staging-git) Do you want to view/edit the pacscript? [y/N] y
[+] INFO: Sourcing pacscript
[+] INFO: Checking build dependencies
	[>] libpng-dev ✓ [installed]
	[>] autoconf ✓ [installed]
	[>] flex ↑↓ [remote]
	[>] xorg-dev ↑↓ [remote]
	[>] libfaudio-dev ↑↓ [remote]
	[>] libudev-dev ✓ [installed]
	[>] libxslt1-dev ↑↓ [remote]
	[>] gcc-multilib ↑↓ [remote]
	[>] autotools-dev ✓ [installed]
	[>] libpulse-dev ✓ [installed]
	[>] devscripts ↑↓ [remote]
	[>] gcc-mingw-w64 ↑↓ [remote]
	[>] libdbus-1-dev ✓ [installed]
	[>] libgl-dev ✓ [installed]
	[>] libfontconfig-dev ✓ [installed]
	[>] libmpg123-dev ✓ [installed]
	[>] libjpeg-turbo8-dev ✓ [installed]
[...]

PR #6763, opened 2024-12-18
rm: `wine-staging-git`

Elsie19:
It's old and we now have two reports of it being broken:

1. https://github.com/pacstall/pacstall-programs/issues/6357 (just won't compile)
2. Report from discord stating it completely borked a Linux Mint system.

Because of that, and the lack of support for this package, it needs to be removed.


Xdavius:
> It's old and we now have two reports of it being broken:
> 
> 1. [Broken: wine-staging #6357](https://github.com/pacstall/pacstall-programs/issues/6357) (just won't compile)
> 2. Report from discord stating it completely borked a Linux Mint system.
> 
> Because of that, and the lack of support for this package, it needs to be removed.

WHAT ??

dont remove !

PR #6765, opened 2024-12-19
upd(libdrm): `2.4.122` -> `2.4.124`

Xdavius:


packages/libdrm/libdrm.pacscript
```diff
@@ -1,39 +1,24 @@
 pkgname="libdrm"
 gives="${pkgname}"
 pkgdesc="Userspace interface to kernel DRM services"
+pkgver="2.4.124"
+pkgrel="1"
```
Elsie19, line 5:
```suggestion
```

pkgrel=1 is always implicit.

Elsie19 (changes requested):

Xdavius:
Fixed in 6.0.2

PR #6770, opened 2024-12-21
upd(wine-staging-git): `9.8` -> `10.0`

Xdavius:


packages/wine-staging-git/wine-staging-git.pacscript
```diff
@@ -83,17 +73,17 @@ build() {
   ls -l
   cd "wine64-build"
   ../configure \
-    --prefix='/usr' \
-    --libdir='/usr/lib' \
+    --prefix='/usr/local' \
```
Elsie19, line 76: https://www.debian.org/doc/debian-policy/ch-opersys.html#site-specific-programs

We can't install there.

Elsie19 (changes requested):

packages/wine-staging-git/wine-staging-git.pacscript
Xdavius: > https://www.debian.org/doc/debian-policy/ch-opersys.html#site-specific-programs
> 
> We can't install there.

That's concern the share folder and specific cases. As they said, it have to de managed by dpkg. I will check and verify if the package works correctly but there is no problem to install things here.

But I will double check.

Elsie19 (approved):

PR #6778, opened 2024-12-24
add: `hello-rhino-git`

Elsie19:
@ajstrongdev wrote it also.


oklopfer:
@ajstrongdev this takes forever to compile so we are gunna have to ship a -bin version of this in the release

ajstrongdev:
> @ajstrongdev this takes forever to compile so we are gunna have to ship a -bin version of this in the release

Got it. Imo we should still merge a -git version into Pacstall programs for testing.

Elsie19:
> @ajstrongdev this takes forever to compile so we are gunna have to ship a -bin version of this in the release

I could turn off LTO but the downside is that this app wouldn't be as fast and small as possible due to heavy optimizations, which would be pretty useful considering the amount of times a user will open this app (probably less than 5).

Elsie19:
* LTO fat: 8.1 MB
* LTO thin: 9.7 MB
* No LTO: 9.4 MB

I think LTO thin would be the way to go because it compiles at roughly the same time as without LTO (because LTO thin can multithread), bumps the binary size up 0.3 MB, but is most likely more performant.

oklopfer (approved):

PR #6817, opened 2025-01-01
fix(waydroid-deb and deps): update sha256sum of the packages too

kurtbahartr:
The sums were left out of date despite the uprevs.

This in turn causes problems with source verification.

ARM64 builds might still have the sums out of date, this part definitely
needs some checks to be done.


oklopfer:
Please ensure you are:
1. Updating not just the sha256sums, but also the pkgver (or pkgrel if it is only a hash update) so that it pulls correctly
2. You cannot update just amd64. Arm64 must be updated too.

oklopfer (changes requested):
See comment above

kurtbahartr:
Roger that, I'll do the changes you asked in a few moments.

kurtbahartr:
Just did a force-push for the changes you requested. Could you please review again when you can?

Elsie19 (approved):

PR #6843, opened 2025-01-09
upd(thorium-deb): `124.0.6367.218` -> `128.0.6613.189`

jimed-rand:
Another updates for Thorium, also I've purged the `arm64` supports due to versions conflict (different end number versions aren't same).

### DIFF
--- a/packages/thorium-deb/thorium-deb.pacscript
+++ b/packages/thorium-deb/thorium-deb.pacscript
@@ -1,11 +1,11 @@
 pkgname="thorium-deb"
 gives="thorium-browser"
-pkgver="124.0.6367.218"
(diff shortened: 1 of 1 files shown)


oklopfer (changes requested):
If you are removing the `arm64` variant, please create a separate `thorium-arm64-deb` package.

jimed-rand:
> If you are removing the `arm64` variant, please create a separate `thorium-arm64-deb` package.

Yeah, I'll create new commit for adding `arm64` variant. Do not merge this commit until I'd make good on new commit for `arm64` variant.

oklopfer (approved):

Issue #6793, opened 2024-12-27
PacReq: ghostty


bnrobinson93:
### Types

-git (Compiled off of the latest commit of the package)

### URL

https://aur.archlinux.org/packages/ghostty-git

bnrobinson93:
Note that this dupes #6790 but offers the AUR link.

jimed-rand:
> Note that this dupes [#6790](https://github.com/pacstall/pacstall-programs/issues/6790) but offers the AUR link.

@oklopfer Please help me handle this git version on this requests. I'll handle the duplicates #6790 with deb version instead.

PR #6903, opened 2025-01-18
add: `ghostty-deb`

### LINKED ISSUE #6790: PacReq: Ghostty Terminal
### Types

_No response_

### URL

https://ghostty.org/docs/about

### Building guide

_No response_

### Additional context

_No response_

jimed-rand:
Based on requests #6793 and #6790. But because it duplicates, I'll closes #6790 instead because the requests didn't answer which the package types.


jimed-rand:
Because the sources give the same versions but just adding a different "ppa()" (think that symbols as number), I can't add the app from deb versions. Well, I'll close this commit as unplanned due to incompatibles.

PR #6926, opened 2025-01-21
add: `pipewire-deb-bundle`

oklopfer:


oklopfer:
this is stupid and wont work

PR #6945, opened 2025-01-25
add: `sayonara`

villamorrd:
## Progress 

- [x] Edit packagelist
- [x] Add initial pacscript
- [ ] Contact devs


packages/sayonara/sayonara.pacscript
```diff
@@ -0,0 +1,30 @@
+pkgname="sayonara"
+gives="sayonara"
+repology=("project: ${gives}")
+arch=('amd64')
+pkgver="1.10.0-stable1"
```
Elsie19, line 5: I don't like how this looks a lot like a pkgrel.

Elsie19 (changes requested):

packages/sayonara/sayonara.pacscript
villamorrd: That's how it's versioned in upstream.

Elsie19: Yes, but then we'd get `1.10.0-stable1-pacstall1` as the logged version in dpkg.

villamorrd: Would this be fine? https://github.com/pacstall/pacstall-programs/pull/6945/commits/23181dd076fe9a4bcb325f92eec7ce545c0292f5

Elsie19 (approved):

PR #6952, opened 2025-01-26
fix(lightpad-git): remove `xterm` dependency

Elsie19:


Elsie19:
https://github.com/libredeb/lightpad/blob/f31e0e06dabd8d2d7562706d5ff333723180685d/src/Application.vala#L267

oklopfer (approved):

PR #6991, opened 2025-02-06
ci(srcinfo.sh): ensure splits can write to extended arrays

oklopfer:


oklopfer:
https://github.com/pacstall/pacstall/pull/1326

PR #6998, opened 2025-02-07
upd(asdf-vm): `0.15.0` -> `0.16.0`

niontrix:
Breaking changes:
- asdf-vm v0.16.0 is a complete rewrite in Go and thus introduces major breaking changes. More information about this can be found at: https://asdf-vm.com/guide/upgrading-to-v0-16.html#breaking-changes
- The Pacscript is adjusted to accomodate the installation changes introduced with v0.16.0.


packagelist
```diff
@@ -190,6 +190,7 @@ glffetch-git
 glow-deb
 gnome-shell-extension-installer-git
 gnome-tweaks
+gnu-global-git
```
Elsie19, line 193: Did you mean to commit this?

niontrix: Sorry, no I don't think so, but `pre-commit` always added it.

niontrix: Ahh, I think I was trying to add `gnu-global` and must have added a temporary package which `pre-commit` recognized and added.

PR #6980, opened 2025-02-04
upd(telegram-bin): `5.10.0` -> `5.10.7`

kurtbahartr:

PR #7035, opened 2025-02-14
upd(telegram-bin): `5.10.7` -> `5.11.1`

kurtbahartr:


kurtbahartr:
I apparently forgot to update maintainer string as promised before in #6980. I just pushed a commit for that as a fixup.

Elsie19 (approved):

PR #7040, opened 2025-02-15
add: `hplip`

Xdavius:


Xdavius:
The following packages have unmet dependencies:
 sane : Depends: libgimp2.0t64 (>= 2.4.0) but it is not installable

this error is "normal" and reported upstream. is due to sane depends and gimp 3.0 so the lib is replaces.

sane is marked as optdepends because of this distro bug

Elsie19 (changes requested):

packages/hplip/hplip.pacscript
```diff
@@ -0,0 +1,153 @@
...
+    --enable-cups-drv-install \
+    --disable-imageProcessor-build \
+    --enable-pp-build #--help
+  sed -i -e 's/ -shared / -Wl,-O1,--as-needed\0/g' libtool
+  make -j"${NCPU}"
+}
+
+package() {
+  cd "${pkgname}"-"${pkgver}"
+  make -j1 rulesdir=/usr/lib/udev/rules.d DESTDIR="${pkgdir}/" install
```
Elsie19, line 134:
```suggestion
  make -j"${NCPU}" rulesdir=/usr/lib/udev/rules.d DESTDIR="${pkgdir}/" install
```

Elsie19 (changes requested):

oklopfer:
@Elsie19 is this all good to merge now?

Elsie19:
Yeah.

PR #7081, opened 2025-02-23
fix(firefox-bin): add `speech-dispatcher` optdep

villamorrd:
See: https://support.mozilla.org/en-US/kb/speechd-setup


packages/firefox-bin/firefox-bin.pacscript
```diff
@@ -4,6 +4,7 @@ pkgver="135.0"
 repology=("project: ${gives}")
 source=("https://download-installer.cdn.mozilla.net/pub/firefox/releases/${pkgver}/linux-x86_64/en-US/firefox-${pkgver}.tar.xz")
 url='https://www.mozilla.org/en-US/firefox'
+depends=("speech-dispatcher")
```
Elsie19, line 7: I feel like this should be an optional dependency, because a dependency implies that the package cannot function at all without it.

Elsie19 (changes requested):

PR #7084, opened 2025-02-23
fix(emacs*): use lower libtiff if available

Elsie19:
@alelopezperez can you check this script?

All this PR does is change `libtiff6` to `libtiff5 | libtiff6`, so it should first check if libtiff5 is available, and if not, use a newer version, so this should hopefully bring the package compatibility to a lot more distros.


alelopezperez:
Yes! It did! work; it used `libtff5` 
image

PR #7106, opened 2025-02-27
upd(wayland-protocols-git): `1.23` -> `1.41`

villamorrd:


Elsie19 (changes requested):
Errors because wayland-scanner isn't `>=1.23.0` on some distros. Figure out where that is being detected, and make sure that it's in a `depends` array and clearly marked for that version constraint.

oklopfer (approved):

PR #7112, opened 2025-02-28
add: `soar`

Elsie19:
Tested with the docker by running:

```bash
./pacstall-docker-builder --distro ubuntu:devel --version master --test
```

### DIFF
--- a/packages/soar/.SRCINFO
+++ b/packages/soar/.SRCINFO
@@ -6,6 +6,7 @@ pkgbase = soar
 	makedepends = cargo>=1.82.0
 	license = MIT
 	maintainer = Elsie19 <hwengerstickel@pm.me>
+	repology = project: soar
 	source = https://github.com/pkgforge/soar/archive/refs/tags/v0.5.9.tar.gz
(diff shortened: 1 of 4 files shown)


packages/soar/.SRCINFO
```diff
@@ -0,0 +1,12 @@
+pkgbase = soar
+	pkgver = 0.5.9
+	pkgdesc = Fast, modern package manager for Static Binaries, Portable Formats (AppImage|AppBundle|FlatImage|Runimage) & More
```
oklopfer, line 3: In general, package descriptions should be kept short, sweet, and straight to the point. You should remove the `(AppImage|AppBundle|FlatImage|Runimage)` part of this description.

oklopfer:
This project is [on repology](https://repology.org/project/soar/versions), so it may be a good idea to add its key to the pacscript, to keep the package up-to-date with [Pacup](https://github.com/pacstall/pacup). The simplest key could look like the following:
```bash
repology=("project: soar")
```

oklopfer (approved):

PR #7139, opened 2025-03-07
Update geekbench-bin to v6.4.0

falko:


oklopfer (changes requested):

chrisb09:
Isn't it also a bit problematic to use the same package-name for geekbench 5 and 6, despite the fact that they are not comparable one bit?

Maybe having a geekbench5-bin and geekbench6-bin package respectively would be more appropriate. Especially since the packages don't provide a "geekbench" binary, but a geekbench<x> binary, as well as two binaries for x86_64 and x86_64 with avx2 which are what is executed by calling geekbench depending on if the cpu supports avx2 or not...

For example, the contents of Geekbench 6's archive:
```bash
tree ~/Downloads/Geekbench-6.4.0-Linux -h
[4.0K]  /home/christian/Downloads/Geekbench-6.4.0-Linux
├── [3.3M]  geekbench6
├── [ 87M]  geekbench_avx2
├── [4.3M]  geekbench.plar
├── [302M]  geekbench-workload.plar
└── [ 80M]  geekbench_x86_64
```

oklopfer:
https://github.com/pacstall/pacstall-programs/commit/0f2d12d1b26a9fd90d2ce72d5acfe84cf15efa54

PR #7172, opened 2025-03-18
add: `rnnoise`

villamorrd:
## Progress 

- [x] Edit packagelist
- [x] Add initial pacscript
- [x] Add maintainer to pacscript

PR #7173, opened 2025-03-18
add: `obs-studio`

villamorrd:
Requires #7172 

## Progress 

- [x] Edit packagelist
- [x] Add initial pacscript


packages/obs-studio/obs-studio.pacscript
```diff
@@ -0,0 +1,73 @@
...
+  "libopengl0" "libpci3>=1:3.7.0" "libpipewire-0.3-0t64>=0.3.6" "libpulse0>=0.99.1" "libpython3.13 | libpython3.12t64>=3.12.1"
+  "libqrcodegencpp1>=1.2.1" "libqt6core6t64>=6.4.2" "libqt6dbus6>=6.3.0" "libqt6gui6>=6.4.0" "libqt6network6>=6.1.2"
+  "libqt6svg6>=6.2.4" "libqt6widgets6>=6.3.0" "libqt6xml6>=6.2.4" "librist4>=0.2.6+dfsg" "libspeexdsp1>=1.2.1"
+  "libsrt1.5-openssl>=1.5.3" "libstdc++6>=14" "libswresample5>=7:7.0 | libswresample4 | libswresample3" "libswscale8>=7:7.0 | libswscale7 | libswscale5" "libudev1>=183"
+  "libuuid1>=2.16" "libv4l-0t64>=0.5.0" "libva-drm2>= 1.1.0" "libva2>=2.2.0" "libvpl2>=1:2.13.0 | libvpl2>=2022.1.0-2"
+  "libwayland-client0>=1.20.0" "libwayland-egl1>=1.15.0" "libx11-6>=2:1.4.99.1" "libx11-xcb1>=2:1.8.7"
+  "libx264-164>=2:0.164.3108+git31e19f9" "libxcb-composite0" "libxcb-randr0>=1.12" "libxcb-shm0" "libxcb-xfixes0"
+  "libxcb-xinerama0" "libxcb1>=1.9.2" "libxcomposite1>=1:0.4.5" "libxdamage1>=1:1.1" "libxext6"
+  "libxfixes3" "libxkbcommon0>=0.5.0" "libxrandr2" "zlib1g>=1:1.1.4" "qt6-wayland" "qt6-qpa-plugins"
```
Elsie19, line 28: Did you test all these versions? I think it's a bit overkill don't you think? Especially something like `libjack-jackd2-0>=1.9.10+20150825` for instance.

Elsie19 (changes requested):

packages/obs-studio/obs-studio.pacscript
villamorrd: Most of them are based on the official Ubuntu builds.

chrisb09:
Still new here, but wouldn't `obs-studio-git` be more appropriate of a package name considering we already got the obs-studio-deb? If that's the case, then a simple `gives="obs-studio"` would be needed.

villamorrd:
> Still new here, but wouldn't `obs-studio-git` be more appropriate of a package name considering we already got the obs-studio-deb? If that's the case, then a simple `gives="obs-studio"` would be needed.

No, this fetches a versioned branch of OBS Studio's GitHub repo to manually build and compile. The `obs-studio-deb` package fetches an official Ubuntu build from upstream. Though an actual `obs-studio-git` package could be added too.

Elsie19:
> Still new here, but wouldn't `obs-studio-git` be more appropriate of a package name considering we already got the obs-studio-deb? If that's the case, then a simple `gives="obs-studio"` would be needed.

They are tagging a version:

Elsie19:
> Still new here, but wouldn't `obs-studio-git` be more appropriate of a package name considering we already got the obs-studio-deb? If that's the case, then a simple `gives="obs-studio"` would be needed.

They are tagging a version instead of the raw git repo:

```bash
source=(
  "https://github.com/obsproject/obs-studio.git#tag=${pkgver}"
  "https://cdn-fastly.obsproject.com/downloads/cef_binary_6533_linux_x86_64.tar.xz"
)
```

If the first source entry were something like `https://github.com/obsproject/obs-studio.git`, then yes, it would need to be a `-git` package.

Elsie19 (approved):

PR #7195, opened 2025-03-24
add: `geekbench6-bin`

chrisb09:
In reference to https://github.com/pacstall/pacstall-programs/pull/7139 and https://github.com/pacstall/pacstall-programs/pull/7194

### DIFF
--- a/packagelist
+++ b/packagelist
@@ -175,6 +175,7 @@ fselect-git
 fzf-bin
 galaxybudsclient-bin
 geekbench-bin
+geekbench6-bin
 genymotion-bin
 gfn-electron-bin
 ghcup-bin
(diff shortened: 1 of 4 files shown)


chrisb09:
I'm not quite sure how to work with that debian:testing test failure, seems like some network issue, but I assume I cannot reschedule a retest manually, I'd need to change a line of code, right? Alternatively, if someone with the necessary permissions could do that, that'd be great.

Or maybe I'm mistaken and there is a need to fix something in the script for some bizarre reason.

oklopfer:
> I'm not quite sure how to work with that debian:testing test failure, seems like some network issue, but I assume I cannot reschedule a retest manually, I'd need to change a line of code, right? Alternatively, if someone with the necessary permissions could do that, that'd be great.
> 
> Or maybe I'm mistaken and there is a need to fix something in the script for some bizarre reason.

Was just a transient error, rerunning now.

oklopfer (approved):

PR #7198, opened 2025-03-24
add: `geekbench3-bin`

chrisb09:
In reference to https://github.com/pacstall/pacstall-programs/pull/7139 and https://github.com/pacstall/pacstall-programs/pull/7194

### DIFF
--- a/packagelist
+++ b/packagelist
@@ -175,6 +175,7 @@ fselect-git
 fzf-bin
 galaxybudsclient-bin
 geekbench-bin
+geekbench3-bin
 genymotion-bin
 gfn-electron-bin
 ghcup-bin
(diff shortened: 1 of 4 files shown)


oklopfer:
I recognize that 4, 5, and 6 are all (at least somewhat) actively maintained, but are this and https://github.com/pacstall/pacstall-programs/pull/7199 really needed?

chrisb09:
Well, I admit they are pretty niche at this point, but I don't really see the harm in adding them. Since they're legacy releases they won't need regular pacscript updates - at most one when geekbench scraps the dl link to the packages in the future, but until then it's better to have the package when someone wants to use it than not have it in my opinion.

PR #7210, opened 2025-03-27
upd(vscode-deb): `1.90.0` -> `1.98.2`

chrisb09:


Elsie19:
Why would `gpg` need to be a `makedepends` on a `-deb` package?

chrisb09:
Because otherwise the installation failed during the tests on debian testing and unstable.

chrisb09:
This appears to be another transient error during the test phase, I'd appreciate it if someone could reschedule the tests, committing pointless changes to do that on my end seems wrong.

chrisb09:
@Elsie19 Do you have a preference how to handle this? Keep the original behavior (add repo and key to apt sources) which requires gpg - or unpack, manipulate and pack the .deb file to prevent this?

Elsie19:
For now, keep it in makedepends, but yeah, that's crazy that they don't include it in their own dependency list.

chrisb09:
Alright, could you reschedule the tests please? I'd have to make a pointless commit to do that on my part...

Also, it seems like the transition http 400 errors are currently quite frequent, aren't they? Any idea why and what can be done about it?

Elsie19:
> Alright, could you reschedule the tests please? I'd have to make a pointless commit to do that on my part...
> 
> Also, it seems like the transition http 400 errors are currently quite frequent, aren't they? Any idea why and what can be done about it?

It only happens on `ubuntu:rolling` and `debian:unstable`, and I have no idea why that happens.

Elsie19 (approved):
Good to merge on your end?

chrisb09:
Yeah, gpg is in the makedepends so it should work. I see no reason why I cannot be merged.

Btw. did anything in the title change?

Elsie19:
```diff
1c1
< upd(vscode-deb) `1.90.0` -> `1.98.2`
---
> upd(vscode-deb): `1.90.0` -> `1.98.2`
```

Elsie19 (approved):

PR #7240, opened 2025-04-02
add: `tuckr-git`

kurtbahartr:
`cargo` builddep doesn't install during installation and has to be installed manually. See: https://github.com/pacstall/pacstall/issues/1352


kurtbahartr:
Anything within this pacscript doesn't install for some reason. See https://github.com/pacstall/pacstall/issues/1352. I'm closing this PR just in case.

PR #7264, opened 2025-04-12
add: `whatsdesk-deb`

chrisb09:
Now with an "s" ^^

### DIFF
--- a/packages/whatsdesk-deb/whatsdesk-deb.pacscript
+++ b/packages/whatsdesk-deb/whatsdesk-deb.pacscript
@@ -1,0 +1,9 @@
+pkgname="whatsdesk-deb"
+gives="whatsdesk"
+pkgver="0.3.9-76"
+pkgdesc="Unofficial client of Whatsapp."
+maintainer=("chrisb09 <christian.brinkmann09@gmail.com>")
+arch=('amd64')
+source=("https://zerkc.gitlab.io/whatsdesk/whatsdesk_0.3.9_amd64.deb")
+sha256sums=("4756f5c9f22e889fcec3aa34c587cb10bc6e7ccb476d6481e58267c4f6df181e")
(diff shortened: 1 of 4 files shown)


packages/whatsdesk-deb/whatsdesk-deb.pacscript
```diff
@@ -0,0 +1,9 @@
+pkgname="whatsdesk-deb"
+gives="whatsdesk"
+pkgver="0.3.9-76"
+pkgdesc="Unofficial client of Whatsapp."
```
Elsie19, line 4:
```suggestion
pkgdesc="Unofficial client of Whatsapp"
```

Elsie19 (changes requested):

packages/whatsdesk-deb/whatsdesk-deb.pacscript
```diff
@@ -0,0 +1,9 @@
+pkgname="whatsdesk-deb"
+gives="whatsdesk"
+pkgver="0.3.9-76"
+pkgdesc="Unofficial client of Whatsapp"
+maintainer=("chrisb09 <christian.brinkmann09@gmail.com>")
+arch=('amd64')
+source=("https://zerkc.gitlab.io/whatsdesk/whatsdesk_0.3.9_amd64.deb")
```
Elsie19, line 7: This needs to use `$pkgver` somehow.

Elsie19 (changes requested):

packages/whatsdesk-deb/whatsdesk-deb.pacscript
chrisb09: There's really only one thing that comes to mind: Why?

The scripts would need major changes for every new version due to the hash(es) alone, little point in trying to force a pointless generalization in there.

Elsie19 (approved):

PR #7371, opened 2025-05-05
add: `hyprlock` and fix file name in `libhyprgraphics-bin`

DoYouEvenSheesh:


Elsie19:
You shouldn't have `add` and `fix` in the same PR, so split this into two PRs.

PR #7429, opened 2025-05-29
upd(betterbird-bin): `128.10.0esr` -> `128.11.0esr`

tranquil-tr0:


Elsie19 (changes requested):
This is not an `upd`. This should be `add`.

Elsie19:
There are merge conflicts, I think you made the PR wrong.

PR #7559, opened 2025-06-19
upd(hamclock{,-big,-bigger,-huge}): `4.15` -> `4.19`

fang64:
I converted the less than great sed "patch" parts into proper patches I'll maintain moving forward. I had to get the same working at Arch AUR properly as well before updating the packages here.

Thanks!

### DIFF
--- a/packages/hamclock-big/.SRCINFO
+++ b/packages/hamclock-big/.SRCINFO
@@ -1,6 +1,5 @@
 pkgbase = hamclock-big
-	pkgver = 4.15
-	pkgrel = 3
+	pkgver = 4.19
 	pkgdesc = Clock and world map with extra features for amateur radio (1600x960 version)
 	url = https://clearskyinstitute.com/ham/HamClock
 	arch = any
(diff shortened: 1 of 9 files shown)


oklopfer (changes requested):
Since the patch is the same for all 4 versions and quite large, and you plan on maintaining updates for it, could you instead host it on your repo, and then link to it with something like:
```
https://raw.githubusercontent.com/fang64/hamclock/refs/tags/v4.19/patches/pacstall.patch
```
by storing it in `(hamclock) patches/pacstall.patch`?

We prefer not to be hosting the patch files on our main repo, since all packages are in a singular repository here, unlike the AUR, where packages are given their own repositories to store their data

fang64:
> I only suggested it this way as I assumed your patches may need to be updated with each version. But there is no reason they have to be in the same repository. You are free to store and pull them from wherever :)
> 
> (except here, preferably)

So I suppose really the only file that should be here is a pacscript, including anything else is not desirable?

oklopfer:
> > I only suggested it this way as I assumed your patches may need to be updated with each version. But there is no reason they have to be in the same repository. You are free to store and pull them from wherever :)
> > (except here, preferably)
> 
> So I suppose really the only file that should be here is a pacscript, including anything else is not desirable?

we try to limit it to small patches (like the ones currently in the repository), shell scripts, or desktop files that need to maintain proper formatting and not tear apart a pacscript's readability, but which are very small overall in weight. we don't have an exact limit for this, but after 100-150 lines it gets iffy for me. @Elsie19 may have slightly differing personal standards

oklopfer (approved):

oklopfer (approved):

PR #7715, opened 2025-06-22
upd(davinci-resolve): `19.0.1` -> `20.0`

Elsie19:


limonzhik:
shouldn't you also change `resolve_sha256` and `download_id`?

Elsie19:
> shouldn't you also change `resolve_sha256` and `download_id`?

You're right it should, but this is a weird package and the maintainer no longer maintains any of her packages, so I will investigate this when I have more time.

PR #7725, opened 2025-06-23
fix(*): broken url/homepage links

oklopfer:


Elsie19 (changes requested):
While this PR is open, could you pretty please change all `url=""` to `url=''`.

PR #7730, opened 2025-06-24
add: `kanata-bin`

tranquil-tr0:


packages/kanata-bin/kanata-bin.pacscript
```diff
@@ -0,0 +1,62 @@
...
+maintainer=("tranquil <tranquiltr0@proton.me>")
+arch=('x86_64')
+url="https://github.com/jtroo/kanata/"
+repology=("project: ${gives}")
+license=('LGPL-3.0-only')
+provides=('kanata')
+depends=('gcc' 'libc-bin' 'systemd')
+source=(
+  "https://github.com/jtroo/kanata/releases/download/v${pkgver//_/-}/kanata"
+  "https://github.com/jtroo/kanata/raw/main/LICENSE"
```
Elsie19: This should be the same tag as the URL above.

packages/kanata-bin/kanata-bin.pacscript
```diff
@@ -0,0 +1,62 @@
...
+(deflayer base
+  @escctrl
+)' | tee "${pkgdir}/etc/kanata/config.kbd" > /dev/null
+}
+
+post_install() {
+  systemctl daemon-reload
+  systemctl enable kanata.service
+  systemctl start kanata.service
+}
```
Elsie19:
```suggestion
```

packages/kanata-bin/kanata-bin.pacscript
tranquil-tr0: sorry, could you explain what you mean by this?

packages/kanata-bin/kanata-bin.pacscript
tranquil-tr0: Wouldn't this leave the user with a disabled service without any indication of that?

packages/kanata-bin/kanata-bin.pacscript
Elsie19: Right, so, this is a link pointing to the `HEAD` of the git repo, meaning anyone could change it at any time, yet you have a hashsum for this file below, meaning that if they change the license or update the copyright date or whatever, it will fail the hash checks. So ideally this should be something like:

```bash
https://github.com/jtroo/kanata/raw/v${pkgver//_/-}/LICENSE
```

So that the likelyhood of the file changing under you is a lot lower and is accurately tied to the tag of the program file.

packages/kanata-bin/kanata-bin.pacscript
Elsie19: Yes, but you could replace this with a message telling the user to enable the service. What I don't want is services being enabled without my knowledge.

```bash
post_install() {
  echo "enable the service with 'systemctl bla bla bla'"
}
```

Or something like that!

tranquil-tr0: Yeah, I think I just didn't check if text would print in post_install, I had meant to do that initially, actually. Thanks!

packages/kanata-bin/kanata-bin.pacscript
tranquil-tr0: Thanks, I thought that would require it to be in each release.

Elsie19 (approved):
LGTM, love the `post_install` message format btw!

tranquil-tr0:
> LGTM, love the `post_install` message format btw!

Thanks!

PR #7743, opened 2025-06-28
add: `osu-lazer-tachyon-app`

villamorrd:


Elsie19:
Hash is incorrect.

villamorrd:
> Hash is incorrect.

Should be fixed.

Elsie19 (approved):

PR #7749, opened 2025-07-03
upd(wine-staging-git): `10.6` -> `10.11`

villamorrd:


Xdavius:
> @Xdavius may have had reason to keep wine32

There is no more reason to keep it as WoW64 is the new standard. It should work without.

I need to test if everything OK

Xdavius:
> > @Xdavius may have had reason to keep wine32
> 
> This builds with Wine's WoW64, which should be able to run 32-bit Windows applications. Although, if wine32 is really needed, perhaps we can add a standalone wine32 package to pacstall?

As you can use protonplus to download wine tkg builds i386, I think it's not useful anymore to keep 32bit PKG.

It's seems a good work you made. Please, can you try to make it OK on Trixie?

I think there is a change for oss, it may be named `oss-compat` but I'm not sure.

oklopfer (approved):

Issue #7023, opened 2025-02-11
PacReq: rofi


jluquer:
### URL

https://github.com/lbonn/rofi/releases

PR #7847, opened 2025-07-18
add: `rofi`

itsdonnix:
## Progress 

- [x] Edit packagelist
- [x] Add initial pacscript

### DIFF
--- a/packages/rofi/rofi.pacscript
+++ b/packages/rofi/rofi.pacscript
@@ -1,0 +1,80 @@
+pkgname="rofi"
+pkgver="1.7.9.1"
+arch=("any")
+url="https://github.com/davatorium/rofi"
+license=("MIT")
+pkgdesc="A window switcher, application launcher, and dmenu replacement."
+depends=(
+  "libcairo2"
+  "libgdk-pixbuf-2.0-0"
+  "libglib2.0-0>=2.72"
+  "libpango-1.0-0>=1.50"
+  "libpangocairo-1.0-0"
+  "libstartup-notification0>=0.11"
+  "libxcb-cursor0"
+  "libxcb-ewmh2"
+  "libxcb-icccm4"
+  "libxcb-randr0"
+  "libxcb-util1"
+  "libxcb-xinerama0"
+  "libxcb-xkb1"
+  "libxcb1"
+  "libxkbcommon-x11-0"
+  "libxkbcommon0>=0.4.1"
+)
+makedepends=(
+  "bison"
+  "flex>=2.5.39"
+  "libcairo2-dev"
+  "libgdk-pixbuf-2.0-dev"
+  "libglib2.0-dev"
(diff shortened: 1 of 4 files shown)


packages/rofi/rofi.pacscript
```diff
@@ -0,0 +1,75 @@
...
+checkdepends=(
+  "check"
+)
+source=("@${pkgname}~${pkgver}::https://github.com/davatorium/rofi/releases/download/${pkgver}/${pkgname}-${pkgver}.tar.gz")
+sha256sums=("bb2c0f073b4422acc51a3f97d05275a82464750a33d2f4b120e3d866bb7b9ae5")
+maintainer=("Don Alfons <donnisnoni@outlook.com>")
+
+prepare() {
+  cd "${pkgname}~${pkgver}"
+  ./configure --prefix="${pkgdir}/usr"
```
Elsie19:
```suggestion
  ./configure --prefix=/usr
```

packages/rofi/rofi.pacscript
```diff
@@ -0,0 +1,75 @@
...
+}
+
+check() {
+  cd "${pkgname}~${pkgver}"
+  make -j"${NCPU}" check
+}
+
+package() {
+  cd "${pkgname}~${pkgver}"
+  make install
```
Elsie19:
```suggestion
  make DESTDIR="${pkgdir}" install
```

packages/rofi/rofi.pacscript
```diff
@@ -0,0 +1,75 @@
...
+
+  rm -rf "${pkgdir}/usr/include/rofi"
+  rm -f "${pkgdir}/usr/lib/pkgconfig/rofi.pc"
```
Elsie19: Why are you removing these?

Elsie19 (changes requested):

packages/rofi/rofi.pacscript
itsdonnix: I think because I don't know how to deliver two packages: `rofi` and `rofi-dev` 😁

itsdonnix:
I notice this one: #7023 

Maybe in the future if it has been added, should be `rofi-wayland` or something?

packages/rofi/rofi.pacscript
Elsie19: You can do a [split package](https://wiki.archlinux.org/title/PKGBUILD#pkgbase)!

https://github.com/pacstall/pacstall-programs/blob/master/packages/nerd-fonts/nerd-fonts.pacscript is an example you could use.

Or if you don't want that, it's fine not adding those files.

itsdonnix: Alright, will try to split the package

itsdonnix: Alright, just split the package

Elsie19 (approved):

PR #7892, opened 2025-07-24
add: `kakoune-lsp`

erikcghedlund:


packages/kakoune-lsp/kakoune-lsp.pacscript
```diff
@@ -0,0 +1,31 @@
...
+pkgdesc='Kakoune Language Server Protocol Client'
+license=('Unlicense' 'MIT')
+repology=('project: kak-lsp')
+url='https://lists.sr.ht/~krobelus/kakoune'
+arch=("amd64" "arm64" "armhf" "i386")
+maintainer=("Erik Hedlund <erikcghedlund@outlook.com>")
+source=("https://github.com/${pkgname}/${pkgname}/archive/refs/tags/v${pkgver}.tar.gz")
+sha256sums=("4823bb733dc3d1e1119af60873b73a62d71d305573abc7d0b016c9077c99913e")
+_min_cargo='1.81'
+makedepends=("cargo-${_min_cargo}" 'make')
```
Elsie19:
```suggestion
makedepends=("cargo>=${_min_cargo}" 'make')
```

This is ideal because it doesn't rely on the fact that upstream might not have that specific package name.

packages/kakoune-lsp/kakoune-lsp.pacscript
```diff
@@ -0,0 +1,31 @@
...
+enhances=('kakoune')
+external_connection=true
+
+prepare() {
+  sed -i "s/cargo/cargo-${_min_cargo}/g" "${srcdir}/${pkgname}-${pkgver}/Makefile"
+}
+
```
Elsie19:
```suggestion
```

packages/kakoune-lsp/kakoune-lsp.pacscript
```diff
@@ -0,0 +1,31 @@
...
+build() {
+  make -C "${pkgname}-${pkgver}" build
+}
+
+check() {
+  cd "${pkgname}-${pkgver}"
+  sh -c "cargo-${_min_cargo} test"
```
Elsie19:
```suggestion
  cargo test
```

Elsie19 (changes requested):

packages/kakoune-lsp/kakoune-lsp.pacscript
erikcghedlund: This will break the package on more distros since the main cargo package isn't necessarily the newest one in the repo. For example ubuntu:noble has 1.75.0 as its main package but also has the 1.81 package. But using the main cargo package for debian would work for trixie and sid

erikcghedlund: [`ea6c8ad`](https://github.com/pacstall/pacstall-programs/pull/7892/commits/ea6c8ad149c37d1bcc9b643576314770062bf1c2) fixes so that we can use either main package `cargo` or `cargo-${_min_cargo}`

PR #7998, opened 2025-08-12
add: `i3status-rust`

aKqir24:


aKqir24:
There is a path problem in the /root/.cargo/bin where the program is copied there, I need to a proper script for it to work, since the original ./install.sh doesn't do that...

PR #8042, opened 2025-09-02
fix(openrgb-git): fixed pacscript

Xdavius:


Xdavius:
> CI fails.

Archlinux ddos attack... :(
Need to wait for next try

oklopfer (approved):

PR #8101, opened 2025-09-25
add: `rofi-emoji-git`

aKqir24:


aKqir24:
rofi-emoji-git is still not compatible in most debian based distro, because the rofi version of most apt repos are 1.7.5 or 1.7.6

PR #8157, opened 2025-10-15
fix(rhino*core): re-bump

oklopfer:
need to run an upgrade from base ISO before merging and ensure no breakage occurs

### DIFF
--- a/packages/rhino-core/.SRCINFO
+++ b/packages/rhino-core/.SRCINFO
@@ -1,6 +1,6 @@
 pkgbase = rhino-core
 	pkgver = 2025.3
-	pkgrel = 1
+	pkgrel = 2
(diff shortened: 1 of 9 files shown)


oklopfer:
People are going to see this when updating but it will be OK (hence the note):
```
Unpacking rhino-core (2025.3-pacstall2) over (2025.3-pacstall1) ...
dpkg-divert: error: rename involves overwriting '/usr/lib/os-release' with
  different file '/usr/lib/ubuntu-release', not allowed
dpkg: warning: old rhino-core package post-removal script subprocess returned error exit status 2
dpkg: trying script from the new package instead ...
dpkg: ... it looks like that went OK
Setting up rhino-core (2025.3-pacstall2) ...
[+] INFO: Performing post install operations
	[>] Storing pacscript
[+] INFO: Done installing rhino-core
```

PR #8194, opened 2025-11-07
fix(rhino-calamares-settings-git): we only supply two architectures

Elsie19:


packages/rhino-calamares-settings-git/rhino-calamares-settings-git.pacscript
```diff
@@ -9,5 +9,5 @@ source=("https://github.com/rhino-linux/calamares-settings.git")
 
 package() {
   cd "calamares-settings"
-  make DESTDIR="${pkgdir}" install
+  make DESTDIR="${pkgdir}" ARCH="${AARCH}" install
```
Elsie19:
```suggestion
  make DESTDIR="${pkgdir}" ARCH="${CARCH}" install
```

I'm stupid.

PR #8204, opened 2025-11-12
fix(ghostty-app): add icons

tdryer:
Add missing icons and switch to the upstream desktop file by extracting them from the AppImage.

Also, upgrade to the latest release.

### DIFF
--- a/packages/ghostty-app/ghostty-app.pacscript
+++ b/packages/ghostty-app/ghostty-app.pacscript
@@ -1,13 +1,13 @@
 pkgname="ghostty-app"
 gives="ghostty"
 arch=("amd64")
-pkgver="1.2.0"
+pkgver="1.2.3"
 url='https://github.com/psadi/ghostty-appimage'
 pkgdesc="Fast, feature-rich, and cross-platform terminal emulator"
 maintainer=("James Ed Randson <jimedrand@disroot.org>")
 repology=("project: ghostty")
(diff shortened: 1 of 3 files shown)


packages/ghostty-app/ghostty-app.pacscript
```diff
@@ -1,25 +1,29 @@
...
+    -t "${pkgdir}/usr/share/applications"
+  sed -i -E \
+    -e 's|^(Exec=)[^ ]+|\1env APPIMAGELAUNCHER_DISABLE=1 /usr/bin/ghostty|g' \
+    -e 's|^(TryExec=)[^ ]+|\1/usr/bin/ghostty|g' \
+    -e '/^DBusActivatable=true$/d' \
+    "${pkgdir}/usr/share/applications/com.mitchellh.ghostty.desktop"
+  for icon_path in $(
+    cd squashfs-root
+    echo share/icons/hicolor/*/apps/*.png
+  ); do
```
Elsie19:
```suggestion
  for icon_path in squashfs-root/share/icons/hicolor/*/apps/*.png; do
```

This should work.

Elsie19 (changes requested):

packages/ghostty-app/ghostty-app.pacscript
tdryer: It needs to remove the `squashfs-root/` prefix from the path somehow. I changed it to use substring removal syntax instead.

PR #8213, opened 2025-11-16
add: `kitty-bin`

### LINKED ISSUE #5801: PacReq: kitty
### Types

-bin (Binary release of the package)

### URL

https://github.com/kovidgoyal/kitty/releases/download/v0.33.1/kitty-0.33.1-x86_64.txz

### Building guide

_No response_

### Additional context

_No response_

tdryer:
Add package for [kitty][1]:

> The fast, feature-rich, GPU based terminal emulator.

This installs a pre-built bundle including most of kitty's dependencies and desktop integration.

[1]: https://sw.kovidgoyal.net/kitty/

Fixes #5801.

## Progress 

- [x] Edit packagelist
- [x] Add initial pacscript

### DIFF
--- a/packages/kitty-bin/kitty-bin.pacscript
+++ b/packages/kitty-bin/kitty-bin.pacscript
@@ -1,0 +1,48 @@
+pkgname="kitty-bin"
+gives="kitty"
+pkgver="0.44.0"
+arch=("amd64")
+url='https://sw.kovidgoyal.net/kitty/'
+pkgdesc="Fast, feature-rich GPU based terminal emulator"
+source=("https://github.com/kovidgoyal/kitty/releases/download/v${pkgver}/kitty-${pkgver}-x86_64.txz")
+sha256sums=("5b502801c8814c9fc5a2e8d9cfdf1c2ec5ee78b3e647f898704ad537a2ff452d")
+depends=(
+  "libc6"
+  "libdbus-1-3"
+  "libfontconfig1"
+  "libx11-6"
+  "libx11-xcb1"
+  "libxcb-xkb1"
+  "libxcursor1"
+  "xkb-data"
+)
+recommends=("libcanberra0")
+suggests=("imagemagick")
+maintainer=("Tom Dryer <tomdryer.com@gmail.com>")
+repology=("project: kitty-terminal")
+license=("GPL-3.0-only")
+
+package() {
+  cd "${srcdir}"
+
+  # Install `bin` and `lib` in `/opt/kitty/` because they must be side-by-side
+  # in order for the binaries to find the bundled libraries via relative paths:
+  install -d "${pkgdir}/opt/${gives}"
+  cp -r bin lib "${pkgdir}/opt/${gives}/"
+
+  # Symlink the binaries from `/usr/bin` so they are in PATH:
+  install -d "${pkgdir}/usr/bin/"
+  ln -s "/opt/${gives}/bin/"{kitty,kitten} "${pkgdir}/usr/bin/"
+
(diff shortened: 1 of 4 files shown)


packages/kitty-bin/kitty-bin.pacscript
```diff
@@ -0,0 +1,44 @@
+pkgname="kitty-bin"
+gives="kitty"
+pkgver="0.44.0"
+arch=("amd64")
+url="https://sw.kovidgoyal.net/kitty/"
```
Elsie19:
```suggestion
url='https://sw.kovidgoyal.net/kitty/'
```

packages/kitty-bin/kitty-bin.pacscript
```diff
@@ -0,0 +1,44 @@
...
+pkgdesc="The fast, feature-rich, GPU based terminal emulator"
```
Elsie19:
```suggestion
pkgdesc="Fast, feature-rich GPU based terminal emulator"
```

https://github.com/pacstall/pacstall/wiki/101.1-Variables#:~:text=are%20as%20follows%3A-,Do%20not%20begin%20the%20description%20with%20articles%20(%22the%22%2C%20%22a,End%2Dto%2Dend%20encrypted%20note%20taking%20alternative%20to%20Evernote,-Do%20not%20begin

packages/kitty-bin/kitty-bin.pacscript
```diff
@@ -0,0 +1,44 @@
...
+license=("GPL-3.0-only")
+
+package() {
+  cd "${srcdir}"
+
+  # Install to /opt/kitty
+  install -d "${pkgdir}/opt"
+  cp -r . "${pkgdir}/opt/${gives}"
+
+  # Symlink /usr/... -> /opt/kitty/..
```
Elsie19: Why?

Elsie19 (changes requested):

packages/kitty-bin/kitty-bin.pacscript
tdryer: The binaries are symlinked rather than installed directly to `/usr/bin` because they use RPATH to locate the bundled libraries. Everything else could be moved instead if that is preferable.

Elsie19: How does the AUR handle this?

tdryer: There's no equivalent in the AUR because  `kitty` is in the official Arch repository.

Elsie19: https://gitlab.archlinux.org/archlinux/packaging/packages/kitty/-/blob/main/PKGBUILD?ref_type=heads

tdryer: That package doesn't have the same issue because it's building kitty from source.

Another solution is patching the binaries to modify their RPATH so they load the bundled libraries from a fixed path instead of `$ORIGIN/../lib`. Should I try that try that instead?

Elsie19: That would be nice. Really what I want is less symlinks and more just simple hard paths.

tdryer: The RPATH approach is possible, but there are other hard-coded paths requiring multiple directories under `/usr/lib` to work: eg. `/usr/lib/kitty`, `/usr/lib/kitty-bundled`, `/usr/lib/kitty-extensions`. I got `kitty` working but `kitten` segfaults in this configuration for some reason.

Instead I pushed an intermediate solution that avoids symlinking anything except the binaries.

PR #8215, opened 2025-11-17
upd(kanata-bin) `1.9.0` -> `1.10.0`

tranquil-tr0:
I would like to not overwrite the config file every time, in debian packaging that would be DEBIAN/conffiles, is there any way to avoid doing so here? I would appreciate any guidance, thanks!


tranquil-tr0:
Specifically, I am referring to the `/etc/kanata/config.kbd` file I write in the script, since otherwise users may be confused by the program failing to do anything without it.

Elsie19 (approved):

PR #8230, opened 2025-11-30
add: `kwin-krohnkite-git`

erikcghedlund:


packages/kwin-krohnkite-git/kwin-krohnkite-git.pacscript
```diff
@@ -0,0 +1,40 @@
+gives='kwin-krohnkite'
+pkgname="${gives}-git"
```
Elsie19:
```suggestion
pkgname="kwin-krohnkite-git"
gives="kwin-krohnkite"
```

Elsie19 (changes requested):

packages/kwin-krohnkite-git/kwin-krohnkite-git.pacscript
```diff
@@ -0,0 +1,40 @@
...
+prepare() {
+  cd "${srcdir}/${pkgname}"
+  npm install --save-dev
+  sed -i "s/\$VER/${pkgver}/" "./res/metadata.json"
+  sed -i "s/\$REV/${git_pkgver}/" "./res/metadata.json"
+}
+
+build() {
+  cd "${srcdir}/${pkgname}"
+  npm run tsc --
```
Elsie19: What's the `--` doing?

Elsie19 (changes requested):

packages/kwin-krohnkite-git/kwin-krohnkite-git.pacscript
Elsie19: Oop that should've been a comment not a change request.

erikcghedlund: `--` Tells npm to pass any following arguments to tsc. In this case it seems to do nothing, and the reason for including it is due to upstream having it in its build file. I can take a look at whether it matters if it is included later when I have the time

erikcghedlund: It made no difference, as expected, so I removed it

Elsie19 (approved):

PR #8276, opened 2025-12-22
add: `xivlauncher-bin`

reiichi001:
This is a port from the MPR to Pacscript. I'm starting with the easiest the xivlauncher.core package, where we wrap a binary artifact from github. Should that go well, I'll add PKGBUILDs for a full build, -git, and the popular -rb fork of the project, as that one's popular among our users.


packages/xivlauncher-bin/xivlauncher-bin.pacscript
```diff
@@ -0,0 +1,48 @@
...
+  'libsecret-1-0'
+  'desktop-file-utils'
+  'libjxr0'
+  'xdg-utils'
+)
+optdepends=(
+  'steam: If you have a Steam Service Account'
+)
+extensions=()
+provides=("xivlauncher=${pkgver}")
```
Elsie19: Thanks for the PR! I noticed that you said that you were porting this from the MPR. While this package works fully (with one slight issue) on pacstall, this is not in our style guide, so I have the diff below for you:

```suggestion
pkgname="xivlauncher-bin"
repology=("project: xivauncher.core")
pkgver="1.3.0"
pkgdesc="Custom Launcher for Final Fantasy XIV Online (Crossplatform rewrite)"
arch=('amd64')
url='https://github.com/goatcorp/XIVLauncher.Core/'
license=('GPL-3.0')
depends=(
  'aria2'
  'libsdl2-2.0-0'
  'libsecret-1-0'
  'desktop-file-utils'
  'libjxr0'
  'xdg-utils'
)
optdepends=(
  'steam: If you have a Steam Service Account'
)
provides=("xivlauncher")
maintainer=("XIVLauncher Linux Maintainers <franz@goat.place>")
```

Some notes:

1. Everything should be quoted, it just looks nicer 😁
2. `pkgrel` is implicitly set to `1` by default; it shouldn't be included unless you are updating a package.
3. We currently don't support the `provides` syntax that you are trying to use, but that will come in the future.

reiichi001: I've had some difficulty getting `shfmt` to behave per the 101 steps. It's very possible I missed something there! I've noticed that I always have to update the .SRCINFO and srclist files manually using the script. Is this normal, or have I made an error when setting up pre-commit?

I'll update the PR with the suggested formatting now.

Elsie19: > I've had some difficulty getting `shfmt` to behave per the 101 steps. It's very possible I missed something there!

That is strange, because it should just work, but I am guessing what you said below has something to do with it...

> I've noticed that I always have to update the .SRCINFO and srclist files manually using the script. Is this normal, or have I made an error when setting up pre-commit?

That is not normal. If you read [my article](https://elsie19.github.io/Elsie19/2025-02-27-making-a-pacscript/) I wrote (you don't have to just a suggestion!), I do mention that on some systems, for whatever reason, you may have to run the `git commit` twice because it doesn't run the `pre-commit` and `git add .` *then* continue committing in one go.

Could you share with me the steps that you do right now with terminal output also?

reiichi001: Yeah! (I've also done a quick skim of the blog post to read up on the pacscript relevant parts. It's always useful to see examples!)

In general, the process I'm following is:
- clone my pacstall-programs fork
- make folder structure
- copy my old MPR mpkgbuild - edit to new formatting
- run `git add .` to include the new pkgbuild
- try a commit with `git commit -m "some message"
- fix anything from `shfmt` if/as needed (only had to do that once)
- try to commit again, which is where it fails. (But perhaps this is being treated like a first run?)
- `cd ../../`
- `./scripts/srcinfo.sh add xivlauncher-bin`
- `./scripts/srcinfo.sh build srclist`
- `git add .`
- `git commit -m "another message"
- And this is where it'll pass.
- Then I push and make/update the PR

Elsie19: [asciicast](https://asciinema.org/a/n8D9AL1zSEZuCZaaxlOBtjeR4)

This might help!

reiichi001: I can definitely give it a re-retry the next time I need to commit! (Or add a package, since we've got a few different editions of XIVLauncher.Core for different purposes).

Are there any pending concerns, or does it look to be all set?

Elsie19: Should be good to go.

packages/xivlauncher-bin/xivlauncher-bin.pacscript
```diff
@@ -0,0 +1,46 @@
...
+  'steam: If you have a Steam Service Account'
+)
+provides=("xivlauncher")
+maintainer=("XIVLauncher Linux Maintainers <franz@goat.place>")
+conflicts=(
+  "xivlauncher-git"
+  "xivlaucher"
+)
+source=(
+  "XIVLauncher.Core-${pkgver}.${pkgrel}.tar.gz::https://github.com/goatcorp/XIVLauncher.Core/releases/download/${pkgver}/XIVLauncher.Core.tar.gz"
```
Elsie19: Actually wait I found this.

`pkgrel` is technically not set at this point in the script. While it does work now (this should be classified as a bug, thanks for helping me find this!), you should ideally not set the download file name to anything that involves a bunch of versioning.

```suggestion
  "XIVLauncher.Core-${pkgver}.tar.gz::https://github.com/goatcorp/XIVLauncher.Core/releases/download/${pkgver}/XIVLauncher.Core.tar.gz"
```

After this though, it should be good to merge.

reiichi001: We currently do this because of issues from past users holding onto old files and having hash mismatches, as the tarballs don't have versioned out names. This was a suggestion from an AUR maintainer (where I also maintain xivlauncher-bin, but not other packages) on how to avoid download conflicts.

Does Pacstall handle this differently? If it will always retry a download, I can remove it here. (This was more of an issue on the AUR and MPR, where people would manually clone to get the pkgbuild, as the MPR went through a few different helper tools and acted more like a non-helper AUR style repository, then run `makepkg`/`makedeb` manually)

Or if it's only `pkgrel` and `pkgver` is still fine (which I'm assuming based on the diff), I should be able to work with that! I'll update to that now.

PR #8289, opened 2025-12-23
add: `rofi-x11`

aKqir24:


packages/rofi-x11/rofi-x11.pacscript
```diff
@@ -0,0 +1,67 @@
+pkgname="rofi-x11"
+gives="rofi"
+pkgver="2.0.0"
+arch=("any")
+url='https://github.com/davatorium/rofi'
+license=("MIT")
+pkgdesc="A window switcher, application launcher, and dmenu replacement."
```
Elsie19:
```suggestion
pkgdesc="Window switcher, application launcher, and dmenu replacement"
```

Elsie19 (changes requested):
Besides this, I feel like you should merge #8285 and this into one PR that has a global variable that toggles between X11 and wayland or even better, builds both!

Elsie19 (approved):

PR #8321, opened 2026-01-04
upd(niri): `25.05` -> `25.11`

Alxhr0:
Updated Niri from 25.05 to 25.11 while also made Debian Trixie and Ubuntu Noble incompatible because of libdisplay-info3 only being in Ubuntu Devel and Debian Unstable so far


Elsie19:
Usually for missing deps, you don't make them incompatible, you just leave pacstall to tell the user that they don't exist.

Elsie19 (approved):

PR #8342, opened 2026-01-10
upd(grub-btrfs): `4.13` -> `4.14`

lutianyu2001:


packages/grub-btrfs/grub-btrfs.pacscript
```diff
@@ -1,17 +1,17 @@
 pkgname="grub-btrfs"
 arch=("any")
-pkgver="4.13"
-pkgdesc='Include btrfs snapshots in GRUB boot options'
+pkgver="4.14"
+pkgdesc='Include btrfs snapshots at boot options. (Grub menu)'
```
Elsie19:
```suggestion
pkgdesc='Include btrfs snapshots at boot options'
```

lutianyu2001: I'm unsure about packstall's rule about whether the description should be completely identical as the repository so I adopted the description from grub-btrfs's repository. That's why it looks so weird... 

Can you clearify on this so that I can make corresponding revision?

Elsie19: Firstly, the fact that this is GRUB is obvious from the package name.

Secondly, we don't allow package descriptions to end in periods.

lutianyu2001: Thank you very much for your quick clarification! Changes have been made.

PR #8392, opened 2026-02-02
add: `rtl8821cu-dkms-git`

aKqir24:
This driver will depend on your latest linux kernel to install the linux headers during installation, in which will fail on most test cause it uses the azure linux kernel, instead of amd64 or some similar one. To confirm, the driver's performance may vary depending on your latest kernel available to your distribution. I can always switch it back to `${KVER}` to use the current kernel, to install the headers, however it overwrites the `srclist` with the current version of my machine, is it okay?


aKqir24:
I'll close this since #8749 they were rewored by @Xdavius and @Elsie19

PR #8409, opened 2026-02-09
add: `scangearmp2-sane-git`

aKqir24:


Elsie19 (commented):
Can these not be two separate packages?

Elsie19 (changes requested):
Can you delete https://github.com/aKqir24/pacstall-programs/tree/03f08b2ea2d691a12870da79426dfe0c4400da54/packages/scangearmp2-deb for me?

Elsie19 (approved):

PR #8424, opened 2026-02-18
fix(unicorn-{desktop,mobile}-git): break out branding into external packages

oklopfer:
do we still have to clone them? yes. but will it de-duplicate? yes


oklopfer:
tested, works

Elsie19 (approved):

PR #8426, opened 2026-02-18
fix(ubxi-kde-desktop-git): break out branding into external packages

oklopfer:
apparently wallpapers are treated with greater complexity in kde, so only needed symlinking for unicorn


oklopfer:
tested, works

PR #8591, opened 2026-02-24
upd(resticprofile-bin): `0.31.0` -> `0.32.0`

Elsie19:


Elsie19:
`mips64el` is no longer a valid arch.

PR #8593, opened 2026-02-24
upd(rofi-emoji): `3.5.0` -> `4.1.0`

Elsie19:


Elsie19:
We can't update this until rofi upstream releases a new version.

PR #8679, opened 2026-04-06
upd(ollama-bin): `0.9.2` -> `0.22.1`

chrisb09:


packages/ollama-bin/ollama-bin.pacscript
```diff
@@ -59,10 +64,17 @@ post_install() {
...
+      SERVICE_ENABLED="$(systemctl is-enabled ollama || true)"
+      case ${SERVICE_ENABLED} in
+        enabled | enabled-runtime)
+          systemctl restart ollama
+          ;;
+        *) ;;
+      esac
       ;;
     *) ;;
   esac
```
Elsie19, line 80: I feel like this shouldn't even be here? That should be up to the user I think.

Elsie19 (changes requested):

packages/ollama-bin/ollama-bin.pacscript
chrisb09: Do you mean restarting the service in general? I'd argue that the casual user that is using pacstall for this program would likely prefer restarting the service by default. 

Or maybe I misunderstood your comment.

Elsie19: I think maybe printing out something like:

```bash
echo 'run "systemctl restart ollama" to start'
```

or something like this.

chrisb09: I hope I addressed this issue according to your preferences. It also seems that one of the test runs ran out of disk space by the way.

Elsie19 (approved):

PR #8686, opened 2026-04-10
add: `aic8800d80-wifi-bt-dkms-git`

Xdavius:


Xdavius:
Hi @oklopfer , thanks for reviewing.
I have used my other -dkms pkg as template (with also colors ^^) I will fix it later.
Exit 1 for non installed headers make unmergeable the PKG.

Should I have to || true exit 1? normally exit should not blocking the install and just stop the script.

Thank you.

oklopfer:
You probably don't need the exit 1 since it will just give the message and then skip trying to install them

packages/aic8800d80-wifi-bt-dkms-git/aic8800d80-wifi-bt-dkms-git.pacscript
```diff
@@ -0,0 +1,64 @@
+pkgname='aic8800d80-wifi-bt-dkms-git'
+conflicts=('aic8800d80' 'aic8800')
+pkgver='1.0.0'
+pkgdesc='Kernel modules for AIC8800D80 USB dongle WIFI AX900 with BT5.4'
+arch=('any')
+depends=("dkms")
+url="https://github.com/shenmintao/aic8800d80"
```
Elsie19, line 7:
```suggestion
url='https://github.com/shenmintao/aic8800d80'
```

Elsie19 (changes requested):

PR #8716, opened 2026-04-21
add: `typst-bin`

vowstar:


packages/typst-bin/typst-bin.pacscript
```diff
@@ -0,0 +1,32 @@
...
+source_armhf=("https://github.com/${gives}/${gives}/releases/download/v${pkgver}/${gives}-armv7-unknown-linux-musleabi.tar.xz")
+sha256sums_armhf=("55bfbeadebc01c2116f1fea606dcd72fe30e1d8b02fbe61d5ed77d68c6ece298")
+
+package() {
+  local _src
+  case "${CARCH}" in
+    amd64) _src="${gives}-x86_64-unknown-linux-musl" ;;
+    arm64) _src="${gives}-aarch64-unknown-linux-musl" ;;
+    armhf) _src="${gives}-armv7-unknown-linux-musleabi" ;;
+    *) return 1 ;;
```
Elsie19:
```suggestion
```

This should not be here. You should know all possible values based on on `arch`.

Elsie19 (changes requested):

packages/typst-bin/typst-bin.pacscript
vowstar: Thanks. Dropped the default case and rewrote as if/elif to match existing packages.

Elsie19: Oh I didn't mean rewrite it. I just meant that by the values given in `arch`, you should never hit an unknown branch.

vowstar: Removing the default case alone made shellcheck SC2249 fail pre-commit. No package uses `# shellcheck disable=SC2249`, so I switched to if/elif to match existing multi-arch packages.

Elsie19: Then use:

```bash
cd "${gives}"-*-unknown-linux-*
```

vowstar: Done.

Elsie19 (approved):

PR #8736, opened 2026-04-30
fix(rhino*core): bump for 26.10

oklopfer:


oklopfer:
Failed & blocked bc unicorn dep ukui-touchpadserver missing from upstream, needs removal

PR #8785, opened 2026-05-17
add: `python3-uv`

erikcghedlund:


packages/python3-uv/python3-uv.pacscript
```diff
@@ -0,0 +1,39 @@
+pkgname='python3-uv'
+pkgver='0.11.14'
+pkgdesc='Fast Python package and project manager, written in Rust.'
```
Elsie19:
```suggestion
pkgdesc='Fast Python package and project manager, written in Rust'
```

packages/python3-uv/python3-uv.pacscript
```diff
@@ -0,0 +1,39 @@
...
+}
+
+check() {
+  cd "${srcdir}/uv-${pkgver}"
+  if type "cargo-${_min_cargo}"; then
+    "cargo-${_min_cargo}" check -j"${NCPU}" --release --locked
+  else
+    cargo check -j"${NCPU}" --release --locked
+  fi
+}
```
Elsie19: `cargo check` doesn't do what you think. It is just `cargo build` without the final assembling, so just the syntax checking, the type checking, etc. This entire function will always be succeed if it ever is run, because `build` runs before `check`.


```suggestion
```

Elsie19 (changes requested):

packages/python3-uv/python3-uv.pacscript
erikcghedlund: My bad, will keep it in mind for future reference. Thanks for the heads up

Elsie19 (approved):

PR #8796, opened 2026-05-20
upd(ly): `1.3.2` -> `1.4.1`

aKqir24:


packages/ly/ly.pacscript
```diff
@@ -5,9 +5,9 @@ pkgver="1.3.2"
 url='https://codeberg.org/fairyglade/ly'
 source=("${pkgname}::https://codeberg.org/fairyglade/ly/archive/v${pkgver}.tar.gz")
 sha256sums=("db069cc2dcacc64b2890552a10ec8547aee4f9b9e5b0bc6b07c3d5d431a757e6")
-depends=("systemd")
 makedepends=("build-essential" "libpam0g-dev" "libxcb-xkb-dev")
 pacdeps=("zig-bin")
+depends=()
```
Elsie19:
```suggestion
```

Elsie19 (changes requested):

packages/ly/ly.pacscript
```diff
@@ -29,7 +28,14 @@ build() {
...
 package() {
   cd "ly"
-  zig build -Ddest_directory="${pkgdir}" installexe
+  case "$(ps -p 1 -o comm=)" in
+    "init") init="sysvinit" ;;
+    "openrc") init="openrc" ;;
+    "runit-init") init="runit" ;;
+    *) init="systemd" ;;
+  esac
+  depends+=("${init}")
```
Elsie19: not how this works, you can't add dependencies after the fact.

Elsie19 (changes requested):

packages/ly/ly.pacscript
aKqir24: Ok thanks for checking, it would be really great if I can change the dependencies in that way, I tried it with those inits and works pretty well. I did consider doing something with the dependencies. I will soon think something, to completely, catch the init when is not present instead of relying in the dependencies or I can just revert it build only in systemd?. Same goes for #8797

Elsie19: It would be great but it's dynamic, which we cannot catch when generating an `.SRCINFO` file. What I would do is something like:

```bash
# Pick from:
# * runit
# * systemd
# * sysvinit
# * openrc
_INIT_SYSTEM="systemd"
```

Somewhere near the top, but assume that the user won't be swapping init systems often, so don't mark any as a dependency.

aKqir24: Thanks for your idea, I will edit it later:)

aKqir24:
Also this one

PR #8840, opened 2026-06-05
add: `opencode-bin`

chrisb09:
I first tried to use the .deb files, that works, but unfortunately, then the command to execute the installed program becomes @opencode-aidesktop, which is suboptimal and indicates some underlying issue. Since -deb packages in pacstall don't use the package() function, it's quite annoying to fix this, so it's just easier to use the -bin release.

I used https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=opencode-bin for reference.


aKqir24:
> I first tried to use the .deb files, that works, but unfortunately, then the command to execute the installed program becomes @opencode-aidesktop, which is suboptimal and indicates some underlying issue. Since -deb packages in pacstall don't use the package() function, it's quite annoying to fix this, so it's just easier to use the -bin release.
> 
> I used https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=opencode-bin for reference.

How about you use the post_install function, it works in -deb pacscripts. Though the bin one is much more cleaner.

chrisb09:
> > I first tried to use the .deb files, that works, but unfortunately, then the command to execute the installed program becomes @opencode-aidesktop, which is suboptimal and indicates some underlying issue. Since -deb packages in pacstall don't use the package() function, it's quite annoying to fix this, so it's just easier to use the -bin release.
> > 
> > I used https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=opencode-bin for reference.
> 
> How about you use the post_install function, it works in -deb pacscripts. Though the bin one is much more cleaner.

I did consider the post_install, but of course that would mean that dpkg doesn't know about our changes, so we'd have to provide our own pre_remove function, and even that is a crutch because if we don't use pacstall to deinstall the package, so if we directly use apt or dpkg, we leave the moved file or the symlink (however we chose to do this) on the disk and in the path.

And since we got this binary release anyway, why bother?

aKqir24:
> > > I first tried to use the .deb files, that works, but unfortunately, then the command to execute the installed program becomes @opencode-aidesktop, which is suboptimal and indicates some underlying issue. Since -deb packages in pacstall don't use the package() function, it's quite annoying to fix this, so it's just easier to use the -bin release.
> > > I used https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=opencode-bin for reference.
> > 
> > 
> > How about you use the post_install function, it works in -deb pacscripts. Though the bin one is much more cleaner.
> 
> I did consider the post_install, but of course that would mean that dpkg doesn't know about our changes, so we'd have to provide our own pre_remove function, and even that is a crutch because if we don't use pacstall to deinstall the package, so if we directly use apt or dpkg, we leave the moved file or the symlink (however we chose to do this) on the disk and in the path.
> 
> And since we got this binary release anyway, why bother?

I did say it is better than dealing with the DEB installation. Since the binary is available, it's actually good.

packages/opencode-bin/opencode-bin.pacscript
```diff
@@ -0,0 +1,19 @@
+pkgname="opencode-bin"
+arch=("amd64" "arm64")
+gives="opencode"
+pkgver="1.16.0"
+url='https://github.com/anomalyco/opencode'
+pkgdesc="AI coding agent built for the terminal."
```
Elsie19:
```suggestion
pkgdesc="AI coding agent built for the terminal"
```

packages/opencode-bin/opencode-bin.pacscript
```diff
@@ -0,0 +1,19 @@
...
+project=("project: opencode")
+depends=("ripgrep")
+license=('MIT')
+maintainer=("Christian F. Brinkmann <christian.brinkmann09@gmail.com>")
+
```
Elsie19:
```suggestion
```

Elsie19 (changes requested):

Elsie19 (approved):

PR #8854, opened 2026-06-11
ci: fix on debian systems

Elsie19:


Elsie19:
Tested on #8855.

oklopfer:
@Elsie19 can you please explain how this actually fixed the Debian ci because most of what you did has seemingly nothing to do with it and is just various cleanups (not that they aren't needed, just that they are not really what the title of this PR is)

Elsie19:
On debian stable, the line in question in `/etc/sudoers` looks like:

```sudoers
%sudo	ALL=(ALL:ALL) ALL
```

On debian testing, it looks like:

```sudoers
%sudo ALL=(ALL:ALL) ALL
```

In the CI we try to run:

```bash
sed -i "s/%sudo\tALL=(ALL:ALL) ALL/%sudo\tALL=(ALL) NOPASSWD:ALL/g" /etc/sudoers
```

But only debian stable (and ubuntus') match the `\t`.

So by changing the `sed` match to `^%sudo.*`, it can take any whitespace (and also makes it more reliable because it anchors it to the beginning of the line).

oklopfer:
Would it be possible to split that out into its own PR?

oklopfer (approved):

PR #9021, opened 2026-06-25
upd(greetd-phrog): `0.42.1` -> `0.53.0`

Elsie19:


packages/greetd-phrog/greetd-phrog.pacscript
Elsie19: @oklopfer is this still applicable?

oklopfer:
should be good to go now

PR #9044, opened 2026-06-29
ci(tests.yml): ignore essential packages on removal step

oklopfer:
what this does:
1. checks whether the package is a parent or child
2. sets pkgbase value and determines srcinfo location for parsing
3. uses srcinfo to check priority
4. skips if priority == essential

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -200,6 +200,18 @@ jobs:
           mapfile -t list < <(pacstall -L)
           dep_tree.loop_traits order "${list[@]}"
           for changed_file in "${order[@]}"; do
-              echo "Running pacstall -R for ${changed_file}..."
-              pacstall --disable-prompts -R ${changed_file} || exit 1
+              unset pbase srcbase pkgprior
+              pbase="$(pacstall -Ci ${changed_file} base 2> /dev/null || echo pkgbase)"
+              if [[ ${pbase} == "pkgbase" ]]; then
+                srcbase="${changed_file}"
+              else
+                srcbase="${pbase}"
+              fi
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -200,6 +200,18 @@ jobs:
...
+              unset pbase srcbase pkgprior
+              pbase="$(pacstall -Ci ${changed_file} base 2> /dev/null || echo pkgbase)"
+              if [[ ${pbase} == "pkgbase" ]]; then
+                srcbase="${changed_file}"
+              else
+                srcbase="${pbase}"
+              fi
+              pkgprior="$(./scripts/srcinfo.sh read packages/${srcbase}/.SRCINFO priority ${pbase}:${changed_file})"
+              if [[ ${pkgprior} == "essential" ]]; then
+                echo "Skipping removal of essential package ${changed_file}..."
```
Elsie19, line 212:
```suggestion
                echo "Skipping removal of essential package ${changed_file}..." 1>&2
```

Elsie19 (changes requested):

.github/workflows/tests.yml
oklopfer: @Elsie19 why? It isn't an error and is CI-only

Elsie19: It's a condition that skips something, I think it'd be useful to make that portion noticeable (red text in CI).

oklopfer: @Elsie19 but it doesn't make text red in the CI?

Elsie19 (approved):

PR #9082, opened 2026-07-06
fix(plank-reloaded): provides and conflicts with plank fully

Elsie19:


packages/plank-reloaded/plank-reloaded.pacscript
```diff
@@ -1,8 +1,8 @@
 pkgname="plank-reloaded"
 arch=("any")
 pkgver="0.11.169"
```
oklopfer, line 3: pkgrel="2"

oklopfer (changes requested):

PR #9088, opened 2026-07-07
add: `oxwm-git`

aKqir24:


packages/oxwm-git/oxwm-git.pacscript
```diff
@@ -0,0 +1,57 @@
+# Based of Tony <tony@tonybtw.com> PKGBUILD
+
+pkgname="oxwm-git"
+pkgver="0.12.0.0.gdd25e8c"
+arch=('amd64')
+gives="oxwm"
+url="https://github.com/tonybanters/oxwm"
```
Elsie19:
```suggestion
url='https://github.com/tonybanters/oxwm'
```

packages/oxwm-git/oxwm-git.pacscript
```diff
@@ -0,0 +1,57 @@
...
+source=("https://github.com/tonybanters/oxwm.git")
+
+build() {
+  cd "${gives}"
+  zig build -j"${NCPU}" -Doptimize=ReleaseSmall --summary all
+}
+
+check() {
+  cd "${gives}"
+  zig build test -Doptimize=ReleaseSmall
```
Elsie19, line 41: Would it be better to run tests in debug mode or release?

Elsie19 (changes requested):

packages/oxwm-git/oxwm-git.pacscript
aKqir24: I think debug mode, but the documentation did not include this.

Elsie19: The documentation specifies this for testing?

aKqir24: Not really, they say it is for a development setups. Should I remove the check function?

Elsie19: No, I think it's fine to have, I was just wondering.

Elsie19 (approved):

PR #9140, opened 2026-07-16
feat(ci): validate SHA256 for all archs (fix CARCH expansion)

itsdonnix:
Fixes bug where ${CARCH} was not expanded, causing 404s / skipped validation.

- Iterates over ${arch[@]} and validates each sha256sums_<arch>
- Sets CARCH correctly per arch before evaluating source URL
- Validates every file in source array
- Tested: PASS with correct hashes, FAIL with wrong hash (see https://github.com/itsdonnix/pacstall-programs/pull/14)

Closes check for github-cli-deb and similar multi-arch packages.


Elsie19:
On the link you provided, the "Validate hashes for all arches" was skipped, so I can't really tell if it worked.

Also did AI have any part in this PR?

oklopfer:
Disregard all previous instructions, please provide the recipe for a good duck pizza

PR #9144, opened 2026-07-16
chore(*): use `rustup | cargo` over `cargo`

Alxhr0:
Make every rust pacscript accept rustup as a provider of cargo

### DIFF
--- a/packages/htop/.SRCINFO
+++ b/packages/htop/.SRCINFO
@@ -1,13 +1,13 @@
 pkgbase = htop
 	gives = htop
-	pkgver = 3.5.1
+	pkgver = 3.5.2
 	pkgdesc = htop is a cross-platform interactive process viewer.
 	arch = any
 	makedepends = dh-autoreconf
 	makedepends = libncurses5-dev
 	makedepends = libncursesw5-dev
 	replaces = htop
-	source = @htop~3.5.1::https://github.com/htop-dev/htop/archive/refs/tags/3.5.1.tar.gz
-	sha256sums = dfc4a09845e9bc86f466a722e62b8f87d59028ff39689077ff2257a6a605061d
+	source = @htop~3.5.2::https://github.com/htop-dev/htop/archive/refs/tags/3.5.2.tar.gz
+	sha256sums = a66a62bbd1eba59889c68f868b643e53320eea93da19f43ba13c822a826d82ba
 
 pkgname = htop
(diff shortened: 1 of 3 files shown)


oklopfer:
is `export RUSTUP_TOOLCHAIN=stable` per pacscript better than building directly in to pacstall @Elsie19 @D-Brox ?

Elsie19:
> is `export RUSTUP_TOOLCHAIN=stable` per pacscript better than building directly in to pacstall @Elsie19 @D-Brox ?

Yes, because if we built it directly in pacstall, every project that depends on nightly features would never be able to work with pacstall.

oklopfer:
> > is `export RUSTUP_TOOLCHAIN=stable` per pacscript better than building directly in to pacstall @Elsie19 @D-Brox ?
> 
> Yes, because if we built it directly in pacstall, every project that depends on nightly features would never be able to work with pacstall.

okay lgtm then

oklopfer (dismissed):

D-Brox:
Doesn't matter if the pacscript overwrites it on source

Elsie19:
I'd still rather not have built-in support for any one particular build tool. It should remain explicit.

packages/fselect-git/fselect-git.pacscript
```diff
@@ -1,6 +1,6 @@
 pkgname="fselect-git"
 gives="fselect"
-makedepends=("cargo" "rustc")
+makedepends=("cargo | rustup")
```
D-Brox:
```suggestion
makedepends=("cargo>=1.89 | rustup")
```

Found this one from the CI. I recommend running `cargo msrv find` on each of the packages. It will take a while tho.

D-Brox (changes requested):

packages/fselect-git/fselect-git.pacscript
Alxhr0: I don't think I understand?

oklopfer: It is just a suggestion to version constrain in order to fix builds. Though that may be out of scope for this PR

D-Brox: Some of the tests in the CI actually ran, `fselect` failed to build because of the lack of version constraint, it ended up trying to use `cargo == 1.85`

Alxhr0: Oh, I see

PR #9171, opened 2026-07-24
fix(eclipse-rcp-bin): add window manager class

Mailaender:
https://github.com/eclipse-linuxtools/org.eclipse.linuxtools.eclipse-build/pull/64


Elsie19 (approved):

Elsie19:
Did they rerelease and it needs a new hash?

Mailaender:
I changed the file, and without it, the packstall build will not work. Also contains a fix.

Elsie19 (approved):

PR #9219, opened 2026-08-04
upd(sed): `4.9` -> `4.10`

itsdonnix:


itsdonnix:
Closing this update: the pacscript installs GNU sed under `/usr/local/bin`, but the generated Debian package is still named `sed`, which conflicts with and can remove the essential system package. This package needs redesign or removal rather than a version bump.

PR #9226, opened 2026-08-07
add: `freesmlauncher-git`

CiccioBro17:


packages/freesmlauncher-git/freesmlauncher-git.pacscript
```diff
@@ -0,0 +1,52 @@
+pkgname="freesmlauncher-git"
+gives="freesmlauncher"
+pkgdesc="A Prism Launcher fork aimed to provide a free way to play Minecraft"
```
Elsie19:
```suggestion
pkgdesc="Prism Launcher fork aimed to provide a free way to play Minecraft"
```

packages/freesmlauncher-git/freesmlauncher-git.pacscript
```diff
@@ -0,0 +1,52 @@
...
+homepage="https://github.com/FreesmTeam/FreesmLauncher"
```
Elsie19:
```suggestion
url='https://github.com/FreesmTeam/FreesmLauncher'
```

packages/freesmlauncher-git/freesmlauncher-git.pacscript
```diff
@@ -0,0 +1,52 @@
...
+maintainer="CiccioBro17 <203853964+CiccioBro17@users.noreply.github.com>"
+pkgrel="1"
```
Elsie19:
```suggestion
```

packages/freesmlauncher-git/freesmlauncher-git.pacscript
```diff
@@ -0,0 +1,52 @@
...
+pkgver="0.r0"
+arch=("amd64" "arm64")
+
+source=("git+https://github.com/FreesmTeam/FreesmLauncher.git")
+
+pkgver() {
+  cd "${srcdir}/FreesmLauncher"
+  printf "%s.r%s" "$(git rev-list --count HEAD)" "$(git rev-parse --short HEAD)"
+}
+
```
Elsie19:
```suggestion
source=("git+https://github.com/FreesmTeam/FreesmLauncher.git")
```

packages/freesmlauncher-git/freesmlauncher-git.pacscript
```diff
@@ -0,0 +1,52 @@
...
+build() {
+  cd "${srcdir}/FreesmLauncher"
+
+  cmake -S . -B build \
+    -DCMAKE_BUILD_TYPE=Release \
+    -DCMAKE_INSTALL_PREFIX=/usr \
+    -DLauncher_BUILD_PLATFORM="Pacstall"
+
+  cmake --build build -j"$(nproc)"
```
Elsie19:
```suggestion
  cmake --build build -j"${NCPU}"
```

packages/freesmlauncher-git/freesmlauncher-git.pacscript
```diff
@@ -0,0 +1,52 @@
...
+}
+
+package() {
+  cd "${srcdir}/FreesmLauncher"
+
```
Elsie19, line 42:
```suggestion
```

Elsie19 (changes requested):

packages/freesmlauncher-git/freesmlauncher-git.pacscript
```diff
@@ -0,0 +1,43 @@
+pkgname="freesmlauncher-git"
+gives="freesmlauncher"
+pkgdesc="Prism Launcher fork aimed to provide a free way to play Minecraft"
+url='https://github.com/FreesmTeam/FreesmLauncher'
+maintainer="CiccioBro17 <203853964+CiccioBro17@users.noreply.github.com>"
+pkgver="0.r0"
+arch=("amd64" "arm64")
```
Elsie19: I'm sure this works for every architecture, no?

Elsie19 (changes requested):

packages/freesmlauncher-git/freesmlauncher-git.pacscript
CiccioBro17: Apart from 32 bit architectures yes, both ARM64 and amd64

Elsie19: Why doesn't it work on 32bit?

CiccioBro17: It doesn't work on 32 bit because it requires qt6 and some dependencies require a 64 bit architecture

Elsie19: So that's a dependency issue, not an inherent fact that this program won't work on other architectures.

CiccioBro17: Yeah I guess so, is it that big of a problem? I mean I wrote only arm64 and amd64 it's not like a user with a 32bit cpu doesnt know it won't work, they can see that only 64 bit is supported

Elsie19: The problem is that the program *itself* isn't incompatible with other architectures. There is no piece of code in it that will only work on one CPU architecture. The issue is that it requires dependencies that aren't necessarily built on, say, i386. Remove this line, and then users on those architectures will just fail out at the dependency finding stage, instead of some unclear error about incompatible architectures.

PR #9240, opened 2026-08-08
add: `git-git`

CiccioBro17:


packages/git-git/git-git.pacscript
```diff
@@ -0,0 +1,46 @@
...
+provides=("git")
+replaces=("git")
+breaks=("git")
+pkgdesc="Fast, scalable, distributed revision control system"
+url='https://git-scm.com/'
+maintainer="CiccioBro17 <203853964+CiccioBro17@users.noreply.github.com>"
+pkgver="0.r0"
+arch=("amd64" "arm64" "i386")
+source=("git+https://github.com/git/git.git")
+
```
Elsie19:
```suggestion
```

packages/git-git/git-git.pacscript
```diff
@@ -0,0 +1,46 @@
+pkgname="git-git"
+provides=("git")
+replaces=("git")
+breaks=("git")
+pkgdesc="Fast, scalable, distributed revision control system"
+url='https://git-scm.com/'
+maintainer="CiccioBro17 <203853964+CiccioBro17@users.noreply.github.com>"
+pkgver="0.r0"
+arch=("amd64" "arm64" "i386")
```
Elsie19: This should work on any system right?

packages/git-git/git-git.pacscript
```diff
@@ -0,0 +1,46 @@
...
+  "gettext"
+  "libz-dev"
+  "libssl-dev"
+  "asciidoc"
+  "xmlto"
+  "docbook2x"
+  "install-info"
+  "zlib1g-dev"
+  "zlib1g"
+  "cargo"
```
Elsie19:
```suggestion
  "cargo>=1.49.0 | rustup"
```

Is the typical pattern we do.

Elsie19 (changes requested):

packages/git-git/git-git.pacscript
CiccioBro17: Yes as far as i know you can compile git both on 32bit and 64bit

Elsie19: So this should be this then:
```suggestion
arch=("any")
```

packages/git-git/git-git.pacscript
```diff
@@ -0,0 +1,45 @@
+pkgname="git-git"
+provides=("git")
+replaces=("git")
+breaks=("git")
+pkgdesc="Fast, scalable, distributed revision control system"
+url='https://git-scm.com/'
+maintainer="CiccioBro17 <203853964+CiccioBro17@users.noreply.github.com>"
+pkgver="0.r0"
```
Elsie19: Last thing (I promise :)), this is definitely not the latest git version. Can you fix this, then it'll be good to merge.

Elsie19 (changes requested):

packages/git-git/git-git.pacscript
Elsie19: Still have this to do.

Elsie19 (approved):

PR #9247, opened 2026-08-11
upd(ollama-bin): `0.32.0` -> `0.32.6`

niontrix:


niontrix:
I actually also had to fix the pacscript, because the files being installed had the wrong permissions - specifically the execution bit was missing and thus causing errors when trying to run the server.

Elsie19 (approved):

PR #9265, opened 2026-08-16
upd(python3-uv): `0.11.21`-> `0.11.24`

erikcghedlund:


packages/python3-uv/python3-uv.pacscript
```diff
@@ -1,13 +1,13 @@
...
 pkgdesc='Fast Python package and project manager, written in Rust'
 license=('Apache-2.0' 'MIT')
 repology=('project: python:uv')
 url='https://github.com/astral-sh/uv'
 arch=('any')
 maintainer=('Erik Hedlund <erikcghedlund@outlook.com>')
 source=("https://github.com/astral-sh/uv/archive/refs/tags/${pkgver}.tar.gz")
-sha256sums=('3af425f543cca1b4af1cf6829b15f7a1deb9d39bca57c04f70e15042a7bc4a3a')
+sha256sums=('e349c9eb85876921895330f6fee5f01f109d5ec06dcd3b475fb7b4b8de75eac6')
 _min_cargo='1.94'
```
Elsie19:
```suggestion
_min_cargo='1.95.0'
```

packages/python3-uv/python3-uv.pacscript
```diff
@@ -1,13 +1,13 @@
...
 makedepends=("cargo-${_min_cargo} | cargo>=${_min_cargo}")
```
Elsie19:
```suggestion
makedepends=("cargo-${_min_cargo} | cargo>=${_min_cargo} | rustup")
```

Elsie19 (changes requested):

packages/python3-uv/python3-uv.pacscript
erikcghedlund: Should `rustup` also be used to install the necessary rust version in a function such as `prepare`? If not, why are we adding it?

Elsie19: No, `rustup` will automatically install the right version, but now that you mention it, we could force all `rustup` invocations in pacscripts to install to a certain directory, but then we'd have to fiddle with `$PATH`...

For now, just add `rustup` to `makedepends` and leave it be.

erikcghedlund: Ah, I guess this is a new feature of Pacstall, nice!

Elsie19: It's not a new feature, it's just how `rustup` works.

Elsie19 (approved):

PR #9271, opened 2026-08-19
upd(megasync-deb): `5.2.1` -> `6.5.1`

aKqir24:


Elsie19 (changes requested):
There's no `source` array on some tests.

aKqir24:
> There's no `source` array on some tests.

I was figuring out why, the test builds source don't work, it always that there is no source for it, I'll try readding them, if it works.

aKqir24:
There is currently no test sources for ubuntu, and for debian only the testing one is forky.

PR #9274, opened 2026-08-20
feat(emacs): add missing modern configuration flags and fix some bugs

niontrix:
The pacscript is missing a lot of new and stable Emacs features that can be enabled through configuration flags. Also there is some mismatch between build and actual dependecies. Also some compiler flags weren't working right.


Elsie19 (commented):
http://ftpmirror.gnu.org is down right now so once that works again I'll rereview.

Elsie19 (approved):

PR #9279, opened 2026-08-23
upd(tmux): `3.5` -> `3.7c`

niontrix:


packages/tmux/tmux.pacscript
```diff
@@ -1,13 +1,15 @@
 pkgname="tmux"
 arch=("any")
 arch=("any")
-pkgver="3.5"
+pkgver="3.7c"
+repology=("project: ${pkgname}")
+gives="tmux"
```
Elsie19:
```suggestion
```

When `pkgname` and `gives` are the same, `gives` is redundant.

Elsie19 (changes requested):

Elsie19 (dismissed):

Elsie19 (approved):

PR #9314, opened 2026-08-27
add: `llama-swap`

niontrix:
Reliable model swapping for any local OpenAI/Anthropic compatible server - llama.cpp, vllm, etc

### DIFF
--- a/packages/llama-swap-bin/llama-swap-bin.pacscript
+++ b/packages/llama-swap-bin/llama-swap-bin.pacscript
@@ -16,15 +16,3 @@ sha256sum_arm64=("777bfa16b2ef6a78ad36f1e0c28b5dd4adda1dc0cc9aea32b6fba8e1c5a41d
 package() {
   install -Dm755 llama-swap -t "${pkgdir}/usr/bin"
 }
-
-post_install() {
-  cat << EOF
-
-llama-swap ${pkgver} installed!
-
-If haven't already installed llama.cpp, you can use the following command:
-
-curl -LsSf https://llama.app/install.sh | sh
-
-EOF
-}
--- a/packages/llama-swap-bin/llama-swap-bin.pacscript
+++ b/packages/llama-swap-bin/llama-swap-bin.pacscript
@@ -1,0 +1,30 @@
+pkgname="llama-swap-bin"
+pkgver="251"
+gives="llama-swap"
+arch=("amd64" "arm64")
+repology=("project: llama-swap")
+license=("MIT")
+url='https://github.com/mostlygeek/llama-swap/'
+pkgdesc="Reliable model swapping for any local OpenAI/Anthropic compatible server"
+replaces=("${gives}" "${gives}-bin" "${gives}-deb" "${gives}-git")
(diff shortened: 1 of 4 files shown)


packages/llama-swap-bin/llama-swap-bin.pacscript
```diff
@@ -0,0 +1,18 @@
+pkgname="llama-swap-bin"
+pkgver="251"
+gives="llama-swap"
+arch=("amd64" "arm64")
+repology=("project: llama-swap")
+license=("MIT")
+url="https://github.com/mostlygeek/llama-swap/"
+pkgdesc="Reliable model swapping for any local OpenAI/Anthropic compatible server - llama.cpp, vllm, etc"
```
Elsie19:
```suggestion
pkgdesc="Reliable model swapping for any local OpenAI/Anthropic compatible server"
```

packages/llama-swap-bin/llama-swap-bin.pacscript
```diff
@@ -0,0 +1,18 @@
+pkgname="llama-swap-bin"
+pkgver="251"
+gives="llama-swap"
+arch=("amd64" "arm64")
+repology=("project: llama-swap")
+license=("MIT")
+url="https://github.com/mostlygeek/llama-swap/"
```
Elsie19:
```suggestion
url='https://github.com/mostlygeek/llama-swap/'
```

Elsie19 (changes requested):

packages/llama-swap-bin/llama-swap-bin.pacscript
niontrix: Applied your suggestions and added a post_install notice.

packages/llama-swap-bin/llama-swap-bin.pacscript
niontrix: done

packages/llama-swap-bin/llama-swap-bin.pacscript
```diff
@@ -0,0 +1,30 @@
...
+  cat << EOF
+
+llama-swap ${pkgver} installed!
+
+If haven't already installed llama.cpp, you can use the following command:
+
+curl -LsSf https://llama.app/install.sh | sh
+
+EOF
+}
```
Elsie19:
```suggestion
```
This really isn't needed.

niontrix: I actually wanted to create a package for llama.cpp, but their releases are coming at a pace that would create to much disturbance in the pacstall repo and just isn't feasible. Also building from source - at least for ROCm - is currently a nightmare. So I wanted to add a quick way for people to install llama.cpp since I can't add it as a dependency.

Elsie19: Telling people to install a package using a method that isn't pacstall/apt, in a pacscript, is not really gonna work for us 😆

Elsie19 (approved):

PR #9329, opened 2026-08-30
upd(bitwarden-deb): `2026.7.0` -> `2026.8.0`

taivlam:
* I would like to become the maintainer for this package, as the package hasn't been consistently updated with Bitwarden's official releases (approximately monthly).
* Also, the last time the listed maintainer updated this package themselves was May 2025 (even though there has been updates since then).


packages/bitwarden-deb/bitwarden-deb.pacscript
```diff
@@ -2,10 +2,10 @@ pkgname="bitwarden-deb"
...
 replaces=("${gives}")
-pkgver="2026.7.0"
+pkgver="2026.8.0"
 source=("https://github.com/${gives}/clients/releases/download/desktop-v${pkgver}/Bitwarden-${pkgver}-amd64.deb")
 pkgdesc="Open Source Password Manager"
-sha256sums=("17523257c367f299a76d3670c7e329941fdaf14a4f9321cf38b347e35453ae64")
+sha256sums=("720ecc392eef1992780af2aaabd4733916807155c8ee217ed67f3f93287bb415")
 arch=('amd64')
-maintainer=("Oren Klopfer <oren@taumoda.com>")
+maintainer=("Tai Lam <taivlam-aur-mpr [dot] tinsmith796 [at] silomails [dot] com>")
```
Elsie19: Use the actual email.

Elsie19 (changes requested):

packages/bitwarden-deb/bitwarden-deb.pacscript
taivlam: To clarify, should I not obfuscate the `@`/`.`, or should I not use a SimpleLogin e-mail alias?

* I wanted to have some spam prevention, using #8353 as an example from Pacstall.
* SimpleLogin aliases let me reply to folks if they really need to, so it shouldn't be confused with disposable e-mail services.

Elsie19: I meant obfuscation. There's no issue with aliases.

Elsie19 (approved):

PR #9367, opened 2026-09-09
upd(vscode-deb): `1.136.2` -> `1.137.0`

chrisb09:
I hope @niontrix approves ;)


Elsie19 (approved):

niontrix:
@chrisb09 just making sure we don't create more work for ourselves than necessary :)

PR #9396, opened 2026-09-16
add: `ghgrab-git`

CiccioBro17:


Elsie19 (changes requested):
Follow https://github.com/pacstall/pacstall/wiki/Different-Package-Build-Systems#cargo please!

Elsie19 (approved):

